PR #109, opened 2026-05-17
chore(fp): start discovery for documenso/documenso

mushgev:
Starting an **fp-discover** run for [`documenso/documenso`](https://github.com/documenso/documenso).

Marked the campaign as `status: discovering` in `docs/fp-automation/campaigns.yaml`. Will:

1. Build truecourse from local source via `pnpm build:dist` (same artifact `publish.yml` ships to npm).
2. Clone documenso at its current default-branch HEAD and run `truecourse analyze --no-llm --no-stash --no-skills`.
3. Triage up to 10 violations per rule (TP/FP/borderline).
4. File one `[fp-fix]` issue per rule with FP rate >= 10% (labels: `fp-fix`, `fp-target:documenso-documenso`).
5. Push baseline numbers into the campaign block on this branch.
6. Post a summary comment here.

This PR is informational — it can be merged at any time.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EBxua95UsFNiSEMLYivxW9)_

### DIFF
--- a/docs/fp-automation/campaigns.yaml
+++ b/docs/fp-automation/campaigns.yaml
@@ -24,7 +24,7 @@ campaigns:
     tech_stack: [typescript, nextjs, prisma]
     priority: 1
     status: discovering
-    baseline: { analyzed_at: null, target_ref: null, total_violations: null, tp: null, fp: null, tp_rate: null }
+    baseline: { analyzed_at: "2026-05-17", target_ref: "8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f", total_violations: 57492, tp: 763, fp: 458, tp_rate: 0.62 }
     final:    { analyzed_at: null, target_ref: null, total_violations: null, tp: null, fp: null, tp_rate: null }
     notes: []
 
--- a/docs/fp-automation/campaigns.yaml
+++ b/docs/fp-automation/campaigns.yaml
@@ -23,7 +23,7 @@ campaigns:
   - target_repo: documenso/documenso
     tech_stack: [typescript, nextjs, prisma]
     priority: 1
-    status: pending
+    status: discovering
     baseline: { analyzed_at: null, target_ref: null, total_violations: null, tp: null, fp: null, tp_rate: null }
     final:    { analyzed_at: null, target_ref: null, total_violations: null, tp: null, fp: null, tp_rate: null }
     notes: []


mushgev:
## Borderline-only rules — no clear FPs, needs human triage

These rules have at least one borderline sample but no clear FPs (FP rate < 10%). Per fp-discover spec, no issue is filed for these; they are listed here for human review:

### `bugs/deterministic/empty-object-type` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/vite-env.d.ts#L3

  Arguments for TP: `type ImportMetaEnv = {}` will match any non-nullish value, which is the precise footgun the rule targets, and using `Record<string, unknown>` is the safer convention. Arguments for FP: this is a Vite-generated ambient type declaration — Vite's own template scaffolds the file with this exact pattern as a placeholder for project-specific env variables, so changing it serves no practical purpose.

### `performance/deterministic/missing-react-memo` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/docs/src/components/ai/page-actions.tsx#L63

  Arguments for TP: ViewOptions is exported and is a reasonable candidate for memoization. Arguments for FP: this is in `apps/docs` and the component receives function props (handlers) which would invalidate memo benefits anyway; memoization is a perf hint, not a correctness issue.

### `performance/deterministic/unbounded-array-growth` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/jobs/client/local.ts#L186

  Arguments for TP: a `slots.push()` inside a loop without an explicit cap is the pattern the rule targets. Arguments for FP: this is a cron-slot generation routine where loop termination is bound by a date/time range (a finite domain), so growth is effectively bounded by the schedule horizon.

### `bugs/deterministic/await-non-thenable` (borderline 2/2)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/client-only/hooks/use-copy-to-clipboard.ts#L54

  The function parameter `value` is typed `CopyValue = Promise<string> | string`, so when given a string the `await` is a no-op (rule's literal intent met). However, when given a Promise the await is necessary, so removing it would break the polymorphic API. Either interpretation is defensible: rule says await-on-non-thenable is wasted, code says we need to handle both branches uniformly.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/client-only/hooks/use-copy-to-clipboard.ts#L23

  Same polymorphic API consideration: `text` is `Promise<string> | string`. For the string case the await has no effect, but it's intentional to handle both Promise and value uniformly. Could argue TP (await on non-thenable in string case) or FP (necessary for Promise case).

### `bugs/deterministic/function-return-type-varies` (borderline 5/9)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/_share+/share.$slug.tsx#L9

  Returning `undefined | MetaDescriptor[]` from a Remix `meta` export is the framework-sanctioned way to opt out of meta tags conditionally. The rule's intent (unpredictable return) is technically met, but the API explicitly supports this shape.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/_unauthenticated+/organisation.invite.$token.tsx#L10`

  The loader uses a discriminated union via `state` literal — that's an intentional design pattern (state machine) rather than 'unpredictable'. But rule treats varied shapes as a smell, and there are still 3+ shapes here. Either reading is reasonable.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/_unauthenticated+/organisation.decline.$token.tsx#L9`

  Discriminated-union return with `state` literal is intentional state-machine pattern (InvalidLink vs Success). Rule's intent (callers need to handle disparate shapes) holds, but the union is explicitly typed and consumable.

### `bugs/deterministic/loose-boolean-expression` (borderline 10/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/trpc/server/admin-router/find-email-domains.ts#L46

  `status` is an optional enum string ('PENDING'|'ACTIVE'|undefined). The rule worries about empty-string/null/0 — but enum types here never produce '', only undefined. Reading literally the rule applies; in practice the check is safe. Either way defensible.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/components/general/folder/folder-grid.tsx#L52

  `parentId: string | null` — `if (parentId)` would treat '' as no-parent. If '' is never a valid id it's fine; if it could be the rule's concern is real. Code style preference more than a bug.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/trpc/server/admin-router/find-admin-organisations.ts#L43

  `query: string | undefined`. Empty string search vs no search are typically equivalent here, but rule treats truthy-coercion on string as a smell. Defensible either way.

### `code-quality/deterministic/complex-type-alias` (borderline 2/3)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/jobs/client/_internal/json.ts#L5

  `JsonPrimitive = string|number|boolean|null|undefined|Date|symbol` is a small, semantically cohesive union representing the leaf types of a JSON-like serializer. Decomposing into smaller named types adds indirection without clarity. Rule fires literally at the 7-member threshold but the union is well-named and atomic.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/types/field-meta.ts#L48

  Union of 7 sibling field-meta variants — each is a named type already, so the alias is essentially a typed registry. Rule fires at 7 but breaking it further would just create another level of indirection.

### `code-quality/deterministic/deep-callback-nesting` (borderline 4/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/trpc/server/envelope-router/set-envelope-fields.ts#L32

  Deep nesting is from chained array.map calls building a request payload — refactoring to named functions would add overhead. Rule's intent (readability) holds; code's pattern (fluent transform) is also idiomatic.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/prisma/index.ts#L30

  The nesting is in Prisma extension/driver config builder — passing factories that return factories is the library's API contract, not arbitrary callback hell.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/document/find-documents.ts#L364

  Nesting is from Kysely query-builder `eb` callbacks which are the library's idiom. Rule's intent (readability) is fair but the depth here is fundamental to building parameterized predicates.

### `security/deterministic/unpredictable-salt-missing` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/auth/hash.ts#L18

  The file uses bcrypt (with SALT_ROUNDS) for actual password hashing — the flagged `hashString` is a sha512 helper used for non-password contents (token deduplication / fingerprinting, judging from the name and module). The rule's intent (no rainbow-table protection on passwords) is met if hashString is fed passwords; otherwise FP. Without auditing every caller it's defensible either way.

### `code-quality/deterministic/missing-boundary-types` (borderline 4/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/_authenticated+/admin+/documents.$id.tsx#L28

  TP reading: even Remix loaders benefit from explicit return types for downstream Route.LoaderArgs inference. FP reading: Remix's framework convention is that loaders' return types are inferred and consumed via Route.ComponentProps; explicit annotation duplicates the contract.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/openpage-api/app/growth/total-signer-conversion/route.ts#L4

  TP reading: exported HTTP handlers cross a public API boundary and benefit from explicit Response return types. FP reading: Next.js Route Handlers use framework-inferred Response typing and idiomatic code omits explicit annotations.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/_authenticated+/t.$teamUrl+/settings._layout.tsx#L21`

  Remix loader — same trade-off: explicit annotation helps consumers but duplicates framework-inferred contract.

### `code-quality/deterministic/regex-empty-alternative` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/types/css-vars.ts#L9

  TP reading: `(rem|px|em|%|pt|)` does contain an empty alternative which matches empty string — rule's literal trigger condition is satisfied. FP reading: the empty alternative is intentional to allow a unitless number (e.g. '8' → 8 with default unit) — the trailing `|` is a deliberate optional-unit pattern. The rule's heuristic correctly identifies the pattern but the author's intent is unambiguous.

### `security/deterministic/unverified-cross-origin-message` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/routes/embed+/playground.tsx#L180

  TP reading: a postMessage listener without origin verification can receive messages from any window, which is a real security smell — embed playground accepts auth tokens. FP reading: this is a developer-facing playground page that only displays event.data for debugging (timestamp + JSON.stringify into a chat log) and does not act on the message — no privilege escalation path.

### `architecture/deterministic/duplicate-import` (borderline 2/4)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/multiselect.tsx#L4

  Namespace import (`import * as React`) + named import (`useEffect`) from same module is a common pattern; consolidating to `import * as React, { useEffect }` is awkward, and many linters allow this. On the other hand, the imports could be combined as `import * as React from 'react'; import { useEffect } from 'react'` is technically duplicate.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/api/v1/implementation.ts#L8

  Line 7 is a side-effect import (`import '...time-zones'`) and line 8 is a named import from the same module. These could be consolidated (named import implicitly executes side effects), so technically duplicate; but some codebases keep the bare-side-effect import deliberately to signal intent.

### `bugs/deterministic/duplicate-import` (borderline 2/4)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/api/v1/implementation.ts#L8

  Side-effect-only import + named import from same module. Technically redundant since named import would execute side effects, but some codebases keep the bare import for documentation.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/multiselect.tsx#L4

  Same as architecture/duplicate-import: namespace import + named import from 'react'. Could be merged, but namespace-plus-named is a recognized pattern; opinions differ on whether it's a 'real' duplicate.

### `code-quality/deterministic/class-prototype-assignment` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/konva/skia-backend.ts#L13

  The assignment is adding `toString()` to skia-canvas's Path2D class, which we don't own — class method definition isn't available. Strictly the rule fires on prototype assignment, but the practical alternative (subclassing Path2D and re-exporting) is much more invasive for a polyfill. Reasonable both ways.

### `bugs/deterministic/inconsistent-return` (borderline 4/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/rate-limit/rate-limit-middleware.ts#L27

  Hono middleware idiomatically returns either a Response (to short-circuit) or void after calling next(); the inconsistent return is structurally real and could confuse readers, but it is the canonical Hono pattern and not a bug. A strict reading flags it; a Hono-aware reading would not.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/server/middleware.ts#L18

  Hono middleware that returns next()/c.redirect() in some branches and falls through after await next() in others. This is the standard Hono pattern but it does mix value/no-value returns, so a generic detector can reasonably flag it.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/email/transports/mailchannels.ts#L47

  Nodemailer transport `send` traditionally takes a callback and returns void; returning `callback(...)` from early branches just uses the return-statement-as-shortcut idiom while the async branch fires-and-forgets. A strict reader flags the mixed shape, but it follows the documented transport contract.

### `bugs/deterministic/stateful-regex` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/pdf/auto-place-fields.ts#L7

  The /g regex is reused across calls to page.findText(). If findText resets lastIndex internally (typical for matchAll-style consumers), no bug. If it uses .exec()/.test() repeatedly without reset, the rule's concern is real. Without seeing findText's implementation, the warning is legitimate but may not manifest.

### `code-quality/deterministic/magic-number` (borderline 4/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/trpc/server/recipient-router/schema.ts#L27

  254 is the RFC-5321 max email length — a well-known constant in this context. Strict readers (the rule) want a named constant; many devs would consider .max(254) on an email schema self-explanatory. Either reading is defensible.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/jobs/client/local.ts#L225

  405 is a well-known HTTP status code (Method Not Allowed). The rule's intent is unclear literals; HTTP status codes are arguably industry constants. Defensible either way.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/server/api/download/download.ts#L177

  404 is a universally-understood HTTP status. The rule wants extraction; conventionally inline status codes are acceptable in route handlers.

### `code-quality/deterministic/selector-parameter` (borderline 1/4)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/openpage-api/lib/cors.ts#L69

  `allowed` is data (a string/array) rather than a true boolean flag, so it doesn't strictly fit the 'selector parameter' pattern. However, the function does branch on its presence/type, so the rule's intent (split-into-two-functions) has some merit.

### `code-quality/deterministic/unused-constructor-result` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/utils/i18n.ts#L72

  The `new Intl.Locale(language)` is constructed solely for its throw-or-not validation side effect inside a try/catch — a legitimate, if unusual, pattern. The rule's intent (unclear intent) is reasonable since a reader might miss this; but the author's intent is also defensible.

### `reliability/deterministic/express-async-no-wrapper` (borderline 5/5)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/server/api/files/routes/get-envelope-item-pdf-by-token.ts#L23

  This is a Hono route, not Express. Hono natively handles async handler rejections via its built-in error pipeline, so the wrapper concern from the Express world doesn't strictly apply. But the rule's intent (unhandled rejections in async route handlers) does have a less-severe analogue here — a reader could argue either way.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/server/router.ts#L125

  Hono app.use, not Express. Hono awaits the handler and routes errors through its onError. Still, the surface pattern matches the rule and a reviewer concerned about portability could justifiably flag.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/server/router.ts#L135

  Hono middleware mounted via app.use; same reasoning as above — Hono handles async rejections natively, but the structural pattern matches the rule.

### `reliability/deterministic/promise-all-no-error-handling` (borderline 10/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/trpc/server/admin-router/find-user-teams.ts#L54

  Inside a tRPC procedure, rejections bubble to the tRPC error middleware (and become HTTP error responses) — so no try/catch is functionally needed. The rule's intent is reasonable in general (unhandled rejections), but in this tRPC context the error path is already covered upstream.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/document/resend-document.ts#L132

  Server-only action called from tRPC mutations; rejections propagate to callers that handle errors. Wrapping each Promise.all in try/catch would just rethrow. Defensible either way.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/lib/server-only/document/delete-document.ts#L195

  Same reasoning: callers handle errors. The Promise.all rejecting on first error is intentional (fail-fast on email delivery).

### `security/deterministic/disabled-auto-escaping` (borderline 2/6)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/root.tsx#L139

  The __html payload is a static string literal of CSS; no user input is interpolated, so no XSS risk in practice. But the rule's intent (highlight every dangerouslySetInnerHTML) is itself defensible because future edits could introduce dynamic content.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/docs/src/components/mdx/mermaid.tsx#L65

  The svg comes from Mermaid's render() which produces controlled SVG, but Mermaid input comes from markdown which could be authored content. If the docs site trusts its markdown sources, it's safe; otherwise a real risk. Both readings defensible.

### `architecture/deterministic/god-service` (borderline 4/4)

- (no file path)

  Pro-TP: 502 files in `apps/remix` is genuinely large and the route/component sprawl is exactly what the rule warns about. Pro-FP: `remix` is the top-level web application package in a monorepo, not a 'service' in the architectural-microservice sense — splitting a Remix app is non-trivial and the rule's threshold (120) treats packages and services as interchangeable.

- (no file path)

  Pro-TP: 402 files in `packages/trpc` reflects a sprawling router with many sub-routers — exactly the 'too many responsibilities' signal. Pro-FP: tRPC routers are conventionally grouped in a single package and the file count maps to the size of the API surface, not architectural responsibility.

- (no file path)

  Pro-TP: a UI primitives package with 135 files is on the edge of what a maintained component library should hold. Pro-FP: a UI library naturally contains one file per component and grows linearly with the design system — flagging this as a 'god service' misclassifies it.

### `code-quality/deterministic/hardcoded-url` (borderline 4/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/email/templates/team-delete.tsx#L18

  Pro-TP: 'https://documenso.com' is hardcoded as a default prop value in an email template that may be reused for previews; centralizing it in config is the rule's intent. Pro-FP: it is an email-template default fallback used when the real `baseUrl` isn't passed in (e.g. for preview rendering); moving every default value into config produces churn without functional gain.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/api/v1/openapi.ts#L14

  Pro-TP: the URL appears inside an API description string and the rule literally matches it. Pro-FP: it is a human-readable description in an OpenAPI doc — these are inherently doc-strings and can't reasonably live in configuration.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/docs/src/components/ai/page-actions.tsx#L94

  Pro-TP: a hardcoded chatgpt.com URL embedded in product code matches the rule. Pro-FP: this URL is a third-party brand integration ('Open in ChatGPT') — it's a stable external endpoint, not deployment-specific configuration.

### `code-quality/deterministic/magic-string` (borderline 7/10)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/components/field/field.tsx#L87

  Pro-TP: 'true' is repeated 5 times as a DOM data-attribute value and extracting to a constant aids consistency. Pro-FP: 'true'/'false' are sentinel DOM string values defined by HTML semantics — promoting them to constants is over-engineering.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/select.tsx#L53

  Pro-TP: 'popper' is repeated and could be extracted. Pro-FP: 'popper' is a literal API value from `@radix-ui/react-select`'s `position` prop — extracting a constant doesn't add safety since it's a third-party union.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/components/dialogs/document-move-to-folder-dialog.tsx#L106`

  Pro-TP: 'default' is repeated 3 times as a toast variant. Pro-FP: 'default' is the literal API value of a discriminated union prop from the design-system toast component; extracting a constant adds indirection without safety.

### `code-quality/deterministic/no-void` (borderline 10/10)

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/components/embed/embed-direct-template-client-page.tsx#L298`

  Pro-TP: the rule literally targets the `void` operator and the snippet uses it. Pro-FP: `void someAsyncCall()` is the standard eslint-recommended pattern to mark a deliberately unhandled promise (e.g. in React event handlers); flagging it as 'confusing' is a style preference.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/components/dialogs/envelopes-bulk-delete-dialog.tsx#L155

  Pro-TP: `void` operator used. Pro-FP: idiomatic 'I am intentionally not awaiting this' in a React onClick handler.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/template-flow/add-template-placeholder-recipients.tsx#L464`

  Pro-TP: `void` operator used. Pro-FP: idiomatic 'I am intentionally not awaiting this' in a callback.

### `code-quality/deterministic/readonly-parameter-types` (borderline 5/10)

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/components/document/document-global-auth-access-select.tsx#L13`

  Pro-TP: `value: string[]` is mutable and the rule's literal intent is met. Pro-FP: the type is part of a public React component prop API where consumers expect `string[]` — switching to `readonly string[]` is a breaking change with no functional benefit.

- `https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/remix/app/components/general/envelope/envelope-drop-zone-wrapper.tsx#L141`

  Pro-TP: `fileRejections: FileRejection[]` could be readonly. Pro-FP: the type is dictated by react-dropzone's callback signature — using `readonly` here doesn't actually prevent the library from mutating.

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/document-dropzone.tsx#L35

  Pro-TP: `fileRejections: FileRejection[]` could be readonly. Pro-FP: dictated by react-dropzone's callback signature; the dropzone signatures are externally defined.

### `performance/deterministic/missing-cleanup-useeffect` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/packages/ui/primitives/document-flow/add-subject.tsx#L134

  Pro-TP: the rule's literal check (does the useEffect callback return a cleanup function?) is not satisfied unconditionally — when `container` is null, no cleanup is returned. Pro-FP: the listener registration is also guarded by `if (container)`, so the conditional return is functionally correct — when nothing is registered, nothing needs cleaning up.

### `security/deterministic/permissive-cors` (borderline 1/1)

- https://github.com/documenso/documenso/blob/8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f/apps/openpage-api/lib/cors.ts#L43

  Pro-TP: the literal line sets Access-Control-Allow-Origin to '*'; the rule's intent is met. Pro-FP: this is inside a generic CORS helper that switches on the configured `origin` value — the `*` branch only runs when the caller explicitly configures wildcard CORS.

---

mushgev:
## Discovery complete

**Target:** `documenso/documenso` @ `8f6be474a9ce44f88a5d175fb3130ca7fe3bbf5f`

### Baseline

| Metric | Value |
| --- | --- |
| Total violations | **57,492** |
| Unique rules with violations | 216 |
| Sampled violations (up to 10 per rule) | 1,354 |
| TP | 763 |
| FP | 458 |
| Borderline | 133 |
| **Baseline TP rate** | **0.62** |

Recorded under `campaigns.documenso/documenso.baseline.*` on this branch.

### Issues filed

**99 issues** with labels `fp-fix` + `fp-target:documenso-documenso` — one per rule with FP rate >= 10%. Range: [#110 → #208](https://github.com/truecourse-ai/truecourse/issues?q=is%3Aissue+label%3Afp-target%3Adocumenso-documenso).

The largest broken rules (100% FP across the sample) include:
- `code-quality/deterministic/unsafe-any-usage` (44,737 violations) — type-resolution gaps on Zod, tRPC, Prisma, useState generics, tanstack-table
- `bugs/deterministic/argument-type-mismatch` (3,915 violations)
- `database/deterministic/unvalidated-external-data` — flags every tRPC handler despite `.input(ZSchema)` validation
- `bugs/deterministic/array-callback-missing-return` — misfires on `Promise.all(arr.map(async))`
- `bugs/deterministic/void-zero-argument` — conflates `void promise()` fire-and-forget with `void 0`
- `reliability/deterministic/unhandled-rejection-no-handler` and `reliability/deterministic/uncaught-exception-no-handler` — fire on every Remix route module instead of just entry points
- `architecture/deterministic/data-layer-depends-on-api` / `data-layer-depends-on-external` — layer classifier mislabels UI components
- `architecture/deterministic/route-without-auth-middleware` — flags `app.use()` middleware setup and tRPC mounts
- `code-quality/deterministic/computed-enum-value` — flags string-literal enums
- `code-quality/deterministic/react-useless-set-state` — confuses local shadowing with state identity
- `architecture/deterministic/dead-module` — graph traversal misses cross-package imports

### Borderline-only rules

30 rules had only borderline samples (no clear FPs); per spec no issues were filed for these — they are listed in [an earlier comment](#issuecomment) for human triage. Notable: `bugs/deterministic/loose-boolean-expression`, `code-quality/deterministic/no-void`, `reliability/deterministic/promise-all-no-error-handling`, `reliability/deterministic/express-async-no-wrapper` (the codebase uses Hono/tRPC, not Express), `code-quality/deterministic/magic-number`, `code-quality/deterministic/hardcoded-url`, `code-quality/deterministic/readonly-parameter-types`, `architecture/deterministic/god-service` (Remix app is a monorepo package, not a microservice), and several others where the rule's intent is debatable for this codebase.

### Next

fp-next-fix will pick up the labelled issues one at a time when this PR (or a previous fp-fix PR) merges. This PR can be merged at any time.

---

PR #472, opened 2026-06-03
Hosted EE storage migration, GitHub App PR gate, and LLM-provider seam

mushgev:
This branch lands one cohesive arc for the hosted enterprise edition: every per-repo store becomes a pluggable async seam (interface + `File*` default + `setXStore` registry + async delegators), so EE can keep **all** state server-side in Postgres + a cloud BlobStore while OSS stays **byte-for-byte file+git**. On top of those seams it adds a webhook-driven **GitHub App PR gate** that gates merges on new drift, and an **LLM-provider seam** (claude CLI by default, AI-SDK providers in EE) with an encrypted Models settings page. Everything is capability-gated, and the OSS import boundary is enforced by an architecture test — OSS source never imports `@truecourse/ee-*` (nor the enterprise-only vendor SDKs).

## Hosted EE storage migration

**Seams (`packages/core`).** `analysis-store`, `verify-store`, `registry`, `project-config`, and `ui-state` each become an interface + `File*` default + `set/get/reset` registry + async delegators; every read/write/list helper now returns a Promise. The file impls wrap the existing sync `fs` calls and keep the `LATEST` mtime cache, so OSS round-trips the same files. New seams:

- **`contract-store`** — directory-oriented `.tc` corpus keyed by a `RepoRef` (opaque `repoKey` + `commitSha`). `FileContractStore` is a no-op broker over the live repo tree (never copies, never hashes), with a `materializesInPlace` flag so callers know whether `dir` is the live read-only repo or an EE temp dir to ingest, plus a `safeResolve` traversal guard.
- **`spec-store`** — document-oriented (`claims.json` / `decisions.json` / `scanState`); the file impl reads/writes the exact IL on-disk paths.
- **`repo-ref`** — CLI derives the commit from `HEAD`; the GitHub App passes `repoKey`/`commitSha` explicitly (the PR head SHA, since a shallow clone's HEAD is unreliable after `checkout -f FETCH_HEAD`).
- **`analyze-lock`** — extracts the `.analyze.lock` `O_EXCL` logic into an injectable `AnalyzeLock` seam keyed by the **storage identity** (not the code dir), so EE can serialize two analyses of the same repo even though each runs on its own throwaway clone. `atomic-write.ts` re-exports for back-compat.

`analyze-core` gains a `codeDir` option separating *where code is read* (git/parse/pipeline) from the *storage identity* used for the lock and stores. `spec-in-process` routes decisions/scan-state/contracts through the new seams via a `withContracts` (load → materialize → cleanup) helper and only ingests when the caller passes an explicit `ref` (EE) — OSS omits it, so nothing is copied. The `flow`/`rules`/`violation-query` read services, `resolveProjectForRequest`, and the spec mutators are now async over the active store. New subpath exports: `@truecourse/core/lib/{verify-store,contract-store,spec-store,repo-ref,analyze-lock}` and `/types/verify-snapshot`.

**Postgres / Blob adapters.** Three new private workspace packages:

- **`@truecourse/ee-db`** — the single Drizzle/Postgres layer. `createEeDb(connectionString)` opens one pool, runs the consolidated migration history on boot (migrations `0000`–`0005`, **16 tables**: `gh_installations`, `gh_repos`, `gh_baselines`, `gh_runs`, `llm_provider_config`, `blobs`, `analyses`, `analysis_history`, `verify_runs`, `verify_history`, `registry`, `repo_config`, `repo_ui_state`, `contract_sets`, `spec_sets`, `extraction_cache`), and opens a **separate `lockPool`** dedicated to `pg_advisory_lock`. Bulky state is keyed by an opaque `repo_key`; snapshots/corpora go to the BlobStore.
- **`@truecourse/ee-storage`** — a cloud-agnostic `BlobStore` (`put/get/delete/exists/list`) with **Azure** (managed-identity default), **S3-compatible** (AWS/MinIO/R2), **Postgres-bytea**, and **filesystem** adapters. `selectBlobStore`/`loadBlobStoreConfig` default to the object-store-free `postgres` deploy.
- **`@truecourse/ee-data-store`** — Postgres+Blob implementations of every core seam. Snapshots go to blobs; the queryable index + append-only history live in Postgres, so `appendHistory` is a single `INSERT` (removing the file impl's read-modify-write race). Contracts are **content-addressed** (one immutable `sha256`-keyed blob per unique `.tc` content + a jsonb manifest), with objects written before the manifest row and a conservative mark-sweep GC (`gcContractObjects`). `pack.ts` centralizes hashing, bounded-concurrency `mapLimit`, deterministic manifest sorting, and `assertSafeRel`/`safeJoin` traversal guards — the arbitrary-file-write defense, since manifest paths are attacker-influenceable via a PR. `PgKvCacheStore` is keyed globally so identical spec content shares one LLM call; `PgAnalyzeLock` uses session-level `pg_advisory_lock` on the dedicated lockPool.

## GitHub App PR gate (`ee/packages/github-app`)

`@truecourse/ee-github-app` is a registry-composed plugin: `registerGithubApp` loads `GITHUB_APP_*`, returns `false` (capability off) when unset so SSO-only deploys are unaffected, and otherwise mounts a **public** HMAC-verified webhook router and a **protected**, workspace-scoped connect router, both under `/api/ee/github`, lighting up the `github-gate` capability.

- **Webhook** verifies `X-Hub-Signature-256` in constant time over the captured raw body, acks 202, and dispatches `installation` / `push` / `pull_request` / `issue_comment`. On each PR open/sync/reopen it fires three independent flows: the **drift gate**, a **spec-scan offer**, and an **infer offer**.
- **Drift gate** clones the base + fetches `refs/pull/<n>/head` (works for forks), verifies head vs base, and posts an authoritative `TrueCourse / drift` Check. `decideGate` is pure: new drift at/above `minSeverity` fails on `blocking` repos (else neutral/advisory); no contracts → neutral, no base → neutral, **auto-defaulted spec conflicts → neutral `unresolved-conflicts`** so an ambiguous spec neither passes nor fails. The run is recorded as the idempotency anchor; the summary comment and up to 25 inline review comments post in isolation so cosmetics can't downgrade the Check.
- **Spec-scan / infer** are CodeRabbit-style checkbox comments whose state lives in the comment body. Checking the box (verified Bot marker + write/maintain/admin editor) clones the head, runs scan→generate / `inferInProcess`, persists spec+contracts **server-side keyed by `(owner/repo, head SHA)`**, and rewrites the comment with a dashboard deep link. Fork PRs get an explanatory comment instead.
- **Read-only repo.** Clones authenticate via a per-command `http.extraheader` (token kept out of the URL, git output, and `.git/config`), strip the embedded auth right after clone, and force-remove temp dirs in `finally`. **No commit-back anywhere** — contracts persist server-side.
- **GateStore** seam with `FileGateStore` (local/dev, atomic temp+rename under `~/.truecourse/github-app/`) and `PostgresGateStore` (hosted, on the shared ee-db). The **protected connect router** (`/status`, `/setup`, `POST/PATCH/DELETE /repos/link` + `/repos/config`, `/repos/:owner/:repo/runs`) is scoped to the caller's WorkOS org. **Resend email** is optional (only when `RESEND_API_KEY` is set).

## LLM-provider seam + Models page

- **`@truecourse/llm`** (OSS) defines the single `LlmTransport` primitive (`complete` → Zod-validated object, `completeText` → prose) with a process registry defaulting to the **CLI transport**, so OSS depends only on a `claude` binary. `cli-transport` owns the `claude --print --output-format json` envelope: `CLAUDE_CODE*` env stripping, timeout/`AbortSignal`, opt-in `--json-schema`, and model-arg building. A pluggable `KvCacheStore` backs the LLM-stage caches with the exact OSS file layout by default.
- core `cli-provider` and **both IL packages** (`spec-consolidator`, `contract-extractor`) delete their own claude-spawn / JSON-envelope / `stripCodeFences` plumbing and route through `getLlmTransport()`; the slice/block caches become async through the cache facade (`model-args.ts` removed from both packages).
- **`@truecourse/ee-llm`** implements the transport on the Vercel AI SDK, mapping a `ProviderConfig` to **Anthropic / OpenAI / Bedrock / Copilot**, driving timeouts via `AbortSignal.any` and retrying once on a fallback model for non-abort errors.
- **ee-server** adds AES-256-GCM `crypto` (key = `sha256(TRUECOURSE_SECRET_KEY)`, `v1:<iv>:<tag>:<ct>` blob, last-4 masking) and a single-row `LlmConfigStore` (key encrypted at rest, masked secret-free view for the UI, key preserved only on same-provider upserts and cleared on a provider switch). It registers `GET`/`PATCH /api/ee/llm/config` (masked view; live `testConfig` before persisting, then installs the new transport), installing the active transport at boot with an env fallback.
- **ee client** ships a **Models** settings page (`/settings/models`, behind `llm-config`) and a **GitHub Connect** page (`/integrations/github`, behind `github-gate`), with the matching nav icons registered in `EeNavSlot`.

## Round-2 hardening

- **`pg_advisory_lock` analyze lock on a dedicated pool.** The minutes-long analyze lock holds its connection on a separate `lockPool` (max 20, 30s connect timeout) so it can't starve the short store queries running inside that same analyze on the main pool. Re-entrant acquire fails fast (would self-deadlock across pooled sessions); a failed unlock **destroys** the connection rather than returning a suspect one (the advisory lock auto-releases on session end). Keyed via `hashtext` into the advisory-lock space (key collisions only reduce throughput, never correctness).
- **Gate neutral on unresolved conflicts.** When a PR head spec scan leaves auto-defaulted conflicts, the gate returns neutral `unresolved-conflicts` (checked **before** the drift diff) and emails the notify list, so an ambiguous spec is neither merged-blocking nor a silent pass.

## Notable new env vars

- **GitHub App** (all four required to enable): `GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_APP_WEBHOOK_SECRET`, `GITHUB_APP_SLUG`; optional `RESEND_API_KEY`, `RESEND_FROM`; `WORKOS_APP_URL` as the dashboard-origin fallback.
- **Hosted storage**: `DATABASE_URL` (enables the Postgres/Blob stores); `BLOB_STORE` (`azure` | `s3` | `postgres` | `fs`, default `postgres`) plus per-adapter `AZURE_STORAGE_ACCOUNT`/`CONTAINER`/`CONNECTION_STRING`, `S3_ENDPOINT`/`REGION`/`BUCKET`/`ACCESS_KEY_ID`/`SECRET_ACCESS_KEY`/`FORCE_PATH_STYLE`, and `BLOB_FS_ROOT`.
- **LLM**: `TRUECOURSE_SECRET_KEY` (≥32 chars, AES key for provider secrets); env fallback `LLM_PROVIDER`/`LLM_MODEL`/`LLM_FALLBACK_MODEL`/`LLM_BASE_URL`, `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`/`COPILOT_API_KEY`|`GITHUB_COPILOT_TOKEN`/`AWS_REGION`/`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`; CLI-binary override `CLAUDE_CODE_BIN`/`TRUECOURSE_CLI_BINARY`.

All are documented (commented) in `.env.example` and `ee/README.md`; design captured in `docs/EE_STORAGE_PLAN.md` and `docs/EE_PHASE6_DESIGN.md`, with `docs/PLAN.md` Phase 44.

## OSS boundary

OSS stays **byte-for-byte file+git** — the `File*` defaults wrap the existing sync `fs` calls and preserve the on-disk layout, mtime cache, and atomic-write conventions; nothing is copied unless an EE caller passes an explicit `ref`. `tests/architecture/ee-import-boundary.test.ts` enforces that no OSS source imports `@truecourse/ee-*`, and a second guard forbids OSS source statically importing enterprise-only vendor SDKs (`ai`, `@ai-sdk/*`, `@aws-sdk/*`, `@azure/*`). EE workspace packages appear only as root **devDeps** for the in-process Postgres store tests.

## Testing

- Build green: **17/17 packages**.
- Full suite green **except 3 pre-existing analyzer layer-classification failures** in an untouched package (unrelated to this change).
- Coverage added across the slices: the core store/transport seams converted to async (incl. a new `analyze-codedir` test pinning the *code ≠ storage key* flow), the OSS CLI envelope + AES-GCM crypto + `LlmConfigStore`, `installEeStores`/`sweepStaleTempDirs`, the GitHub App suite (`tests/github-app`, signature/config/`decideGate`/comments/email/file+Postgres stores, webhook/connect/offer/gate integration, and `driftsForCommit` against a real `PgBlobContractStore`+`PgSpecStore`), and pglite-backed coverage of the Pg/Blob analysis, verify, contract (dedupe/GC/traversal-safety), spec, config/ui-state/registry, KV-cache, advisory-lock, and blob stores.


mushgev:
Merged latest `main` (which now includes #473's `@truecourse/shared/llm` transport). #473 and this branch both added an `LlmTransport`, so I reconciled them in merge commit 38b816bc: adopted #473's transport as canonical (cli + `agent` mailbox), repurposed this branch's `@truecourse/llm` to the LLM-stage **cache** seam only, and reshaped `@truecourse/ee-llm` into an `LlmTransport` backend installed process-wide via a new `setDefaultTransport()`. OSS stays byte-identical; the Models page + `agent`/`--io` flags are both preserved. Build 17/17, suite green (bar the 3 pre-existing analyzer failures).

PR #475, opened 2026-06-03
fix(spec): treat total extraction failure as an error, not success (#474)

### LINKED ISSUE #474: spec scan reports success (exit 0, "No open conflicts") when 100% of blocks fail extraction — e.g. expired claude login
### Summary

### Environment

### Steps to reproduce

### Actual behavior

```
◇  docs        27
◇  claims      0
◇  resolved    0
◇  decided     0
◇  open        0
```

- Exit code: **0 (success)**

### Expected behavior

### Root cause

### Suggested fixes

DoilLee2222:
Fixes #474.

## Problem

`truecourse spec scan` shells out to the user's `claude` CLI per block and collects per-block failures as non-fatal. That's right for a few stragglers, but when **every** block fails (e.g. an expired Claude login) the command printed `(N failures)` yet still showed the success outro, suggested `contracts generate`, and **exited 0** — hiding the real, actionable error.

## Changes

- **New classifier** in `@truecourse/spec-consolidator` — `summarizeExtractionFailures` / `isLikelyAuthFailure`: collapses duplicate messages into samples-with-counts, flags a total wipeout (`allFailed`), and detects auth signatures (401, `/login`, invalid key, expired token…) by majority rule.
- **`runSpecScan`** now:
  - surfaces the most common distinct failure messages (instead of just a count),
  - prints a re-authentication hint when failures look like auth,
  - exits `1` when every block fails, and exits `1` on a hard pipeline error (previously exit `0`),
  - gates the `contracts generate` suggestion on `claims > 0` ("No claims extracted — nothing to generate yet.").
- Exit-code / outro / hint policy extracted into a pure, unit-tested `decideScanOutcome`.

## Tests

- 10 unit tests for the failure classifier (auth detection, majority rule, sample collapsing, `allFailed`).
- 6 unit tests for the scan-outcome policy.
- Full suite: **4210 passing**, build clean.

### DIFF
--- a/apps/dashboard/server/package.json
+++ b/apps/dashboard/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.0",
+  "version": "0.6.0-next.11",
   "private": true,
   "type": "module",
   "scripts": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/core",
-  "version": "0.6.0",
+  "version": "0.6.0-next.11",
   "private": true,
   "type": "module",
   "exports": {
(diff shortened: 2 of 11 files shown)


DoilLee2222:
Superseded by #485, recreated as a same-repo PR (now that I have write access on the repo). Continuing review there.

PR #485, opened 2026-06-04
fix(spec): treat total extraction failure as an error, not success (#474)

### LINKED ISSUE #474: spec scan reports success (exit 0, "No open conflicts") when 100% of blocks fail extraction — e.g. expired claude login

DoilLee2222:
Fixes #474. Supersedes #475 (recreated as a same-repo PR).

## Problem

`truecourse spec scan` shells out to the user's `claude` CLI per block and collects per-block failures as non-fatal. That's right for a few stragglers, but when **every** block fails (e.g. an expired Claude login) the command printed `(N failures)` yet still showed the success outro, suggested `contracts generate`, and **exited 0** — hiding the real, actionable error.

## Changes

- **New classifier** in `@truecourse/spec-consolidator` — `summarizeExtractionFailures` / `isLikelyAuthFailure`: collapses duplicate messages into samples-with-counts, flags a total wipeout (`allFailed`), and detects auth signatures (401, `/login`, invalid key, expired token…) by majority rule.
- **`runSpecScan`** now:
  - surfaces the most common distinct failure messages (instead of just a count),
  - prints a re-authentication hint when failures look like auth,
  - exits `1` when every block fails, and exits `1` on a hard pipeline error (previously exit `0`),
  - gates the `contracts generate` suggestion on `claims > 0` ("No claims extracted — nothing to generate yet.").
- Exit-code / outro / hint policy extracted into a pure, unit-tested `decideScanOutcome`.
- Adds an up-front `claude` CLI auth preflight so a scan fails fast (before fanning out hundreds of extraction subprocesses) when the login is missing/expired.

## Tests

- 10 unit tests for the failure classifier (auth detection, majority rule, sample collapsing, `allFailed`).
- 6 unit tests for the scan-outcome policy.
- Added unit tests for the `claude` probe classifier and preflight failure messaging.


packages/spec-consolidator/src/extractor.ts
```diff
@@ -122,6 +122,95 @@ export async function extractClaims(
...
+// success.
+// ---------------------------------------------------------------------------
+
+/**
+ * Patterns that mark a `claude` subprocess failure as an authentication
+ * problem (expired/missing login, invalid key) rather than a transient
+ * error. Matched case-insensitively against the failure message, which for
+ * subprocess exits is `claude exited <code>: <stderr>`.
+ */
+const AUTH_FAILURE_PATTERNS: RegExp[] = [
```
mushgev: Did you fix so the same check will be applied to all commands that use Claude and not only Scan. If yes this code shouldn't be here.

packages/spec-consolidator/src/extractor.ts
```diff
@@ -122,6 +122,71 @@ export async function extractClaims(
...
+//
+// Auth is deliberately *not* sniffed here: every command that spawns `claude`
+// runs the same up-front auth preflight (tools/cli/src/lib/claude-preflight.ts
+// → @truecourse/core/lib/cli-binary), so a broken/expired login is caught
+// before any block runs. Re-guessing it from per-block stderr afterwards would
+// duplicate that check (scan-only) and be far less reliable than the live
+// round-trip the preflight already does.
+// ---------------------------------------------------------------------------
+
+export interface FailureSample {
```
mushgev: I still dont get why we have these changes in this file? and maybe I'm wrong but can you double check if these changes here are legit and if yes reply here why.

DoilLee2222: The behavior itself is needed because we need to distinguish between "a repo that genuinely has no extracted claims" and "a scan where every Claude invocation failed (e.g. expired auth)."
Without that distinction, a 100% extraction failure can be reported as a successful empty scan, which is the issue this PR is addressing.
That said, you're right that this logic is currently only consumed by the scan CLI path. It is not being applied across all Claude-backed commands, so it does not belong in the shared extractor layer. I'll move the failure summarization/auth-classification logic into the CLI layer and keep the extractor focused on returning raw extraction results.

README.md
```diff
@@ -356,13 +356,16 @@ TrueCourse talks to Claude Code via the `claude` CLI. You can tune how that inte
 For packaged installs (`npx truecourse` or `npm install -g truecourse`), the simplest place to set them is `~/.truecourse/.env`. The file is loaded automatically on every invocation:
 
 ```
-CLAUDE_CODE_BINARY=claude             # override the `claude` binary on PATH
+CLAUDE_CODE_BINARY=claude             # override the `claude` binary on PATH (CLAUDE_CODE_BIN also accepted)
 CLAUDE_CODE_MODEL=                    # Claude Code --model flag (empty = default)
 CLAUDE_CODE_TIMEOUT_MS=120000         # per-call timeout (ms)
 CLAUDE_CODE_MAX_RETRIES=2             # retry attempts on parse/validation failure
 CLAUDE_CODE_MAX_CONCURRENCY=10        # max concurrent `claude` processes per run
+TRUECOURSE_SKIP_CLAUDE_CHECK=         # set to 1 to skip the up-front `claude` login preflight
```
mushgev: Don't you think this is overkill? Like why people might want to disable it?

DoilLee2222: it's actually overkill. I think users won't even need that. I will remove it.

PR #497, opened 2026-06-05
chore(drift-fp): contracts store for feast-dev/feast @ 276b6df

mushgev:
⚠️ **DO NOT MERGE** — storage branch for the `feast-dev/feast` drift campaign; deleted automatically at campaign close by `drift-fp-campaign-close`.

---

## Contracts store: `feast-dev/feast`

| Field | Value |
|-------|-------|
| `target_ref` | `276b6df562e16fefba7efb493736ff32046d4a76` |
| Generated at | 2026-06-05 |
| Tool version | truecourse 0.6.2 |
| Transport | `agent` (LLM answered prompts in-session) |

### Doc scope

```
docs/getting-started/concepts
docs/getting-started/components
docs/reference
docs/specs
```

139 docs scanned, 742 claims extracted, 37 conflicts resolved (all accepted defaults — generic heading collisions across 130+ reference docs).

### Artifact counts

**59 .tc contracts:**
- ArchitectureDecision: 2
- AuthRequirement: 2 (OIDC Bearer, Kubernetes RBAC)
- NamedConstant: 15 (auth type strings, env var names, config defaults, port defaults)
- UnenforceableObligation: 40 (narrative overview content)

`contracts validate` — 59 artifacts, no issues.

### Notes

feast is a Python feature store SDK, so the behavioral docs are primarily configuration reference (online/offline store types, compute engines, auth schemes) rather than HTTP API specs. The spec surface yields:
- Auth type constants: `"no_auth"`, `"oidc"`, `"kubernetes"` (verifiable against Python auth module)
- Env var names: `FEAST_OIDC_TOKEN`, `LOCAL_K8S_TOKEN`
- Config defaults: `online_write_batch_size=null`, `pull_latest_features=false`
- Port defaults: remote offline store 8815, remote registry 6570
- CLI flag names: `--skip-source-validation`, `--skip-feature-view-validation`

The probe found NamedConstant 189, Operation 71, Enum 20 — the current extraction is conservative (59 contracts). The `discover` run will test whether these map cleanly to Python code.

cc @mushgev

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHVcjjd7J7iYEHW8x2LQWE)_

PR #500, opened 2026-06-05
chore(drift-fp): start discovery for feast-dev/feast

mushgev:
Starting a drift discovery run for `feast-dev/feast`.

- Storage branch: `claude/drift-fp-store/feast-dev-feast` (PR #497)
- Target ref: `276b6df562e16fefba7efb493736ff32046d4a76`
- Code dir: `.`
- Contracts generated: 59 `.tc` files (ArchitectureDecision ×2, AuthRequirement ×2, NamedConstant ×15, UnenforceableObligation ×40)

This PR marks the campaign `discovering`. The routine will run `truecourse verify` against the pinned contracts, triage each drift-kind for TP/FP, file `[drift-fp-fix]` issues for any FP drift-kinds, and update the baseline in `campaigns.yaml`. Additional commits will follow on this branch as the discovery progresses.

cc @mushgev

---
_Generated by [Claude Code](https://claude.ai/code/session_018uAfRTGCXvBE4jMuV33DgS)_

### DIFF
--- a/docs/drift-fp-automation/campaigns.yaml
+++ b/docs/drift-fp-automation/campaigns.yaml
@@ -66,7 +66,7 @@ campaigns:
     code_dir: "."
     doc_scope: []        # fill before generating (keep the behavioral docs/ .md; drop the 93 .rst)
     priority: 2
-    status: pending
+    status: discovering
     baseline: { verified_at: null, target_ref: null, total_drifts: null, tp: null, fp: null, info: null, fp_rate: null }
     final:    { verified_at: null, target_ref: null, total_drifts: null, tp: null, fp: null, info: null, fp_rate: null }
     notes: ["Probe histogram: NamedConstant 189, Operation 71, Enum 20, Formula 6, Pagination present. Has 93 .rst (invisible to the tool) alongside .md."]
--- a/docs/drift-fp-automation/campaigns.yaml
+++ b/docs/drift-fp-automation/campaigns.yaml
@@ -67,7 +67,7 @@ campaigns:
     doc_scope: []        # fill before generating (keep the behavioral docs/ .md; drop the 93 .rst)
     priority: 2
     status: discovering
-    baseline: { verified_at: null, target_ref: null, total_drifts: null, tp: null, fp: null, info: null, fp_rate: null }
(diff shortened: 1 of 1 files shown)


mushgev:
## Discovery complete — feast-dev/feast

**Target ref:** `276b6df562e16fefba7efb493736ff32046d4a76`

### Issues filed

| Issue | Drift-kind | FP count |
|-------|-----------|---------|
| #501 | `named-constant/no-code-counterpart` | 13 |

### Baseline

| Field | Value |
|-------|-------|
| `total_drifts` | 17 |
| `tp` | 0 |
| `fp` | 13 |
| `info` | 4 |
| `fp_rate` | 1.00 |

All 17 drifts are wrong:
- **13 FPs** (`named-constant/no-code-counterpart`, severity `info`): The named-constant comparator cannot match feast's Python constants to spec identities. Three distinct failure mechanisms — (A) name-normalization mismatch between hierarchical spec identities (`auth.type.kubernetes`) and flat Python names (`KUBERNETES`); (B) the last-segment fallback is gated on `window-global` shape and silently skips Python `const-literal` constants; (C) Pydantic class-body field defaults and decorator/`os.getenv` string literals are not extracted by `extractPyConstantsFromFile`. All 13 confirmed FPs by direct code inspection. Fix filed at #501.
- **4 in `info` bucket** (contract-quality — see below): `architecture.data-store.unmet-choice` × 2 and `architecture.data-store.forbidden-alternative` × 2 at `critical` severity. Not filed as `drift-fp-fix` issues; noted as contract-quality.

### contract-quality: notes

The two `ArchitectureDecision` contracts (`arch.feature-store.components`, `arch.permissions.enforcement-boundary`) were generated using a `decision` prose field that the lifter ignores. Because the lifter defaults to `category = 'data-store'` and `chosen = ''` when those fields are absent, the comparator fires spurious `data-store.unmet-choice` and `data-store.forbidden-alternative` drifts against postgres/mysql/mongodb imports in the codebase.

These contracts should have been `UnenforceableObligation` artifacts (prose descriptions of component structure and permission enforcement boundaries) rather than `architecture-decision` artifacts, since neither doc asserts a technology choice that can be verified. There is also a secondary hardening opportunity: the `compareArchitectureDecision` comparator should guard against `chosen = ''` (return early / return inconclusive) rather than treating an empty chosen value as a testable assertion.

These 4 drifts are excluded from the `fp` count and `fp_rate` because the root cause is contract generation quality, not the verifier comparator.

---

PR #523, opened 2026-06-06
fix(cli): treat soft contract validation as warnings

### LINKED ISSUE #492: `contracts generate` exits 1 with "No contracts were written" on soft validation issues (unresolved cross-references) — but contracts were actually written
## Summary

`truecourse contracts generate` aborts with **exit 1** and prints **“No contracts were written”** when the only validation problems are **soft** issues (unresolved cross‑references). This contradicts the orchestrator, which classifies unresolved cross‑references as non‑blocking and **actually writes the contracts to disk** before the CLI bails. The result is a hard failure + a false “nothing was written” message, even though `.truecourse/contracts/` was populated.

A secondary cosmetic bug compounds it: the same unresolved reference is printed **twice, byte‑for‑byte identical**, because each usage site is reported separately and the message omits the usage location.

Hit while testing `truecourse@0.6.0-next.12` (PR #485 build).

## Environment

- `truecourse@0.6.0-next.12` (also reproduces on current `main` — the logic is unchanged since #337)
- macOS, `claude` CLI v2.1.163

## Repro

```
$ npx truecourse@next spec scan          # → 21 claims, 1 open conflict
$ npx truecourse@next spec resolve --all-defaults
$ npx truecourse@next contracts generate
```

## Observed output

```
┌  Contracts
│
◇  `claude` CLI ready
│
◇  extracted  1/5  .truecourse/specs/claims.json#_shared/forbidden :: _shared → forbidden
◇  extracted  2/5  .truecourse/specs/claims.json#_shared/auth :: _shared → auth
◇  extracted  3/5  .truecourse/specs/claims.json#_shared/data :: _shared → data
◇  extracted  4/5  .truecourse/specs/claims.json#_shared/effects :: _shared → effects
◇  extracted  5/5  .truecourse/specs/claims.json#_shared/overview :: _shared → overview
│
■  Validation gate failed (2 issues):
  ErrorEnvelope:error.envelope.standard: cross-reference ErrorEnvelope:error.envelope.standard doesn't resolve
  ErrorEnvelope:error.envelope.standard: cross-reference ErrorEnvelope:error.envelope.standard doesn't resolve
│
└  No contracts were written. Edit the spec or re-run after fixing.
```

(Exit code 1.)

## Expected

The two issues are **soft** (an `ErrorEnvelope` referenced under an identity no slice declared — exactly the "two slices coined different identities for the same artifact" case the code is designed to tolerate). The orchestrator already wrote every resolvable contract to `.truecourse/contracts/`. So the CLI should:

1. **Not** exit 1 on soft‑only issues.
2. Surface the soft issues as **warnings**, not a fatal "validation gate failed".
3. Report what was actually written (`result.write.written`), not "No contracts were written".
4. Print each distinct unresolved reference once, with its usage site.

### 1. CLI gates on *all* validation issues, ignoring severity — contradicting the orchestrator

The validator marks unresolved cross‑references as `severity: 'soft'` and only treats parse errors / duplicate identities as `'hard'`:

- `packages/contract-extractor/src/validator.ts:88-97` — unresolved refs pushed with `severity: 'soft'`.
- `packages/contract-extractor/src/validator.ts:103-105` — `ok = !hasHard` (soft issues keep the result `ok`).

The orchestrator honours this: on soft‑only issues it drops nothing, calls `writeContracts(...)`, and **returns with both `write.written` populated and `validationIssues` carrying the soft issues** (intended as informational):

- `packages/contract-extractor/src/index.ts:231-260` — only `hard` artifact issues are dropped; only a `resolver`‑kind hard error blocks the write; otherwise it writes.
- `packages/contract-extractor/src/index.ts:266-268` — comment: *"Surface the diagnostics even on a successful write."*

The intent is pinned by a test:

- `tests/contract-extractor/orchestrator.test.ts:118-123` — *"Successful runs may surface SOFT validation issues … those don't block the write. Only HARD issues are a failure signal."* asserts `result.write.written.length > 0`.

But the CLI ignores `severity` entirely and treats any non‑empty `validationIssues` as fatal:

```ts
// tools/cli/src/commands/contracts.ts:114-122
// Surface validation issues — these block the write.
if (result.validationIssues.length > 0) {
  p.log.error(`Validation gate failed (${result.validationIssues.length} issue…):`);
  for (const issue of result.validationIssues) {
    console.log(`  ${issue.artifactKey}: ${issue.message}`);
  }
  p.outro("No contracts were written. Edit the spec or re-run after fixing.");
  process.exit(1);
}
```

So the CLI prints “No contracts were written” and exits 1 even though the orchestrator wrote them. The CLI and orchestrator have disagreed on this since they both landed in #337; the bug only surfaces when an unresolved cross‑reference survives the repair pass. The **dashboard** path (same orchestrator, no CLI gate) is unaffected — it would write the contracts.

### 2. Duplicate, indistinguishable messages

The resolver collects **every reference occurrence** across all files (keyed by usage site, not deduped by identity):

- `packages/contract-verifier/src/resolver/index.ts:182-200` + `collectRefs` at `:445-463`.

So `ErrorEnvelope:error.envelope.standard` referenced from two places → two `unresolvedRefs` entries. The generate‑path validator emits one issue per entry **without** the `usedAt` location:

- `packages/contract-extractor/src/validator.ts:91-96` — message is just `cross-reference <type>:<identity> doesn't resolve`.

→ two byte‑identical lines. Note `truecourse contracts validate` does the opposite — it includes `usedAt.filePath:lineStart` (`tools/cli/src/commands/contracts.ts:307-309`), so the two commands report the same corpus inconsistently.

## Impact

- `contracts generate` exits 1 and tells the user nothing was written whenever the LLM produces an unresolved cross‑ref the repair pass can't fix — a common, expected outcome for real specs.
- This **blocks the documented pipeline** (`spec scan` → `spec resolve` → `contracts generate` → `verify`) and any CI/script keying off the exit code, even though usable `.tc` files were produced.
- The "No contracts were written" message is simply false in this case (files exist in `.truecourse/contracts/`), which is misleading when debugging.

## Suggested fix

In `tools/cli/src/commands/contracts.ts`, partition `result.validationIssues` by `severity`:

- Block (exit 1, "No contracts were written") **only** when the write didn't happen — i.e. there are hard `resolver` issues, or equivalently `result.write.written.length === 0` and `dryRun` is false.
- Otherwise report soft issues (and dropped hard artifacts) as **warnings**, then continue to the normal success summary that lists `result.write.written`.
- De‑duplicate the soft‑issue lines and/or include the usage site so identical refs collapse or become distinguishable (consider adding `usedAt` to the generate‑path message to match `contracts validate`).

Happy to send a PR if this analysis looks right.

lihouwenbin:
## Summary

Fixes #492.

`contracts generate` currently exits 1 whenever `result.validationIssues` is non-empty. That contradicts the contract extractor's severity model: unresolved cross-references are `soft` diagnostics, and the orchestrator still writes resolvable contracts when there are no hard validation issues.

This PR updates the CLI to:

- split validation issues into hard vs soft
- keep hard issues blocking with the existing failure message
- report soft issues as warnings and continue to the normal success path
- de-duplicate identical soft validation messages so repeated unresolved refs do not print byte-for-byte duplicates

## Validation

- `npx pnpm@9.15.0 exec turbo build --filter=@truecourse/contract-extractor --filter=truecourse`
- `PATH=/Users/macbookair/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./node_modules/.bin/vitest run tests/cli/contracts-generate.test.ts tests/contract-extractor/orchestrator.test.ts`

Note: local `corepack pnpm` failed on this machine due to a Corepack npm signature key mismatch, so I used `npx pnpm@9.15.0` matching the repo's packageManager.

### DIFF
--- a/tests/cli/contracts-generate.test.ts
+++ b/tests/cli/contracts-generate.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from 'vitest';
+import type { ValidationIssue } from '../../packages/contract-extractor/src/validator';
+import { partitionValidationIssues } from '../../tools/cli/src/commands/contracts';
+
+describe('contracts generate validation issues', () => {
+  it('keeps soft validation issues non-blocking and de-duplicated', () => {
+    const issues: ValidationIssue[] = [
+      {
+        artifactKey: 'ErrorEnvelope:error.envelope.standard',
+        message: "cross-reference ErrorEnvelope:error.envelope.standard doesn't resolve",
+        severity: 'soft',
+      },
+      {
+        artifactKey: 'ErrorEnvelope:error.envelope.standard',
+        message: "cross-reference ErrorEnvelope:error.envelope.standard doesn't resolve",
+        severity: 'soft',
+      },
+    ];
+
+    const { hardIssues, softIssues } = partitionValidationIssues(issues);
+
+    expect(hardIssues).toEqual([]);
+    expect(softIssues).toHaveLength(1);
+    expect(softIssues[0].severity).toBe('soft');
+  });
+
+  it('keeps hard validation issues blocking', () => {
+    const issues: ValidationIssue[] = [
+      {
+        artifactKey: 'resolver',
+        message: 'duplicate identity Operation:GET /api/orders',
+        severity: 'hard',
+      },
+      {
+        artifactKey: 'ErrorEnvelope:error.envelope.standard',
+        message: "cross-reference ErrorEnvelope:error.envelope.standard doesn't resolve",
+        severity: 'soft',
+      },
+    ];
+
+    const { hardIssues, softIssues } = partitionValidationIssues(issues);
+
+    expect(hardIssues).toHaveLength(1);
+    expect(hardIssues[0].artifactKey).toBe('resolver');
+    expect(softIssues).toHaveLength(1);
+  });
+});
--- a/tools/cli/src/commands/contracts.ts
+++ b/tools/cli/src/commands/contracts.ts
@@ -8,6 +8,7 @@ import {
   hasCanonicalSpec,
   spawnRunner,
 } from "@truecourse/contract-extractor";
+import type { ValidationIssue } from "@truecourse/contract-extractor/validator";
 import { agentTransport } from "@truecourse/shared/llm";
 import { stampGeneratedMarker } from "@truecourse/core/commands/spec-in-process";
 import { trackEvent, bucketFileCount, bucketDuration } from "@truecourse/core/services/telemetry";
@@ -24,6 +25,30 @@ export interface RunContractsGenerateOptions {
   llm?: "cli" | "agent";
   /** I/O dir for the `agent` transport's request/response mailbox. */
   io?: string;
+}
+
+export function partitionValidationIssues(issues: ValidationIssue[]): {
+  hardIssues: ValidationIssue[];
+  softIssues: ValidationIssue[];
+} {
+  const seenSoft = new Set<string>();
+  const hardIssues: ValidationIssue[] = [];
+  const softIssues: ValidationIssue[] = [];
+
+  for (const issue of issues) {
+    if (issue.severity === "hard") {
+      hardIssues.push(issue);
+      continue;
+    }
+
+    const key = `${issue.artifactKey}\0${issue.message}`;
+    if (!seenSoft.has(key)) {
+      softIssues.push(issue);
+      seenSoft.add(key);
+    }
+  }
+
+  return { hardIssues, softIssues };
 }
 
 export async function runContractsGenerate(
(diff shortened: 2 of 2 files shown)


DoilLee2222:
Thanks for the contribution! This PR addresses a real bug — soft validation issues should never block the run.

However, this has already been resolved in `main`. The current implementation goes a bit further than what's proposed here:

- Instead of a `Set` for deduplication, `main` uses a `Map` to collapse repeated messages and display an occurrence count (e.g. `(×3)`)
- The block condition is tighter: we only exit 1 when **zero artifacts were produced** and issues exist, rather than exiting on any hard issue even when other artifacts wrote successfully
- The logic is inlined with comments explaining the severity model, keeping the code easier to follow without an extra exported helper

Since `main` already covers the fix and improves on the approach, we'll close this PR to avoid regressing the existing behavior. Thanks again for catching this and putting together a clean fix with tests!

PR #524, opened 2026-06-06
fix(cli): fail spec scan when extraction fully fails

### LINKED ISSUE #474: spec scan reports success (exit 0, "No open conflicts") when 100% of blocks fail extraction — e.g. expired claude login

lihouwenbin:
## Summary

Fixes #474.

`spec scan` currently prints a success outro when every extraction block fails and no claims are produced. In that case the absence of open conflicts is not evidence that the docs are clean; it means the scan produced no usable claims.

This PR updates the CLI scan path to:

- treat `blocksAttempted > 0`, `failures === blocksAttempted`, and `claims.length === 0` as a fatal extraction failure
- print representative failure locations and errors so auth/subprocess failures are visible
- avoid suggesting `contracts generate` when the scan produced zero claims
- preserve partial-failure behavior when at least some claims are extracted

## Validation

- `npx pnpm@9.15.0 exec turbo build --filter=truecourse`
- `PATH=/Users/macbookair/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH ./node_modules/.bin/vitest run tests/cli/spec-scan.test.ts tests/core/spec-telemetry.test.ts`

Note: local `corepack pnpm` fails on this machine due to a Corepack npm signature key mismatch, so I used `npx pnpm@9.15.0` matching the repo's packageManager.

### DIFF
--- a/tests/cli/spec-scan.test.ts
+++ b/tests/cli/spec-scan.test.ts
@@ -1,0 +1,43 @@
+import { describe, expect, it } from 'vitest';
+import {
+  representativeExtractionFailures,
+  scanSuccessOutro,
+  totalExtractionFailure,
+} from '../../tools/cli/src/commands/spec';
+
+describe('spec scan extraction failures', () => {
+  it('treats all-block extraction failure as fatal', () => {
+    expect(totalExtractionFailure({
+      blocksAttempted: 2,
+      claims: [],
+      failures: [
+        { block: { filePath: 'docs/a.md', headingPath: ['A'] }, error: 'claude exited 1: login expired' },
+        { block: { filePath: 'docs/b.md', headingPath: ['B'] }, error: 'claude exited 1: login expired' },
+      ],
+    })).toBe(true);
+  });
+
+  it('does not treat partial extraction failure as fatal', () => {
+    expect(totalExtractionFailure({
+      blocksAttempted: 2,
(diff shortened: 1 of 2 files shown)


DoilLee2222:
## Closing — superseded by #485

This PR fixes the right problem, but the fix has already landed via #485 (commit `629a0e86`), which was merged earlier and covers everything here — and more.

**What #485 added that this PR doesn't:**
- Auth failure detection with a re-login hint printed to the user
- Preflight `claude` auth check *before* the scan fans out (fails fast instead of burning through all blocks)
- Deduplication + repeat counts on failure messages (not just "up to 3 raw lines")
- `decideScanOutcome()` — a pure, unit-tested function that owns the exit-code/outro policy

The current `main` already handles the total-extraction-failure → exit 1 case, suppresses the `contracts generate` suggestion when no claims were produced, and surfaces failure details. There's nothing left for this PR to add.

Thanks for the work — the diagnosis and approach were correct. Closing as superseded.

PR #543, opened 2026-06-08
fix(verify): stop the stash from sweeping away .truecourse/ contracts (#542)

### LINKED ISSUE #542: verify: stash sweeps away untracked .truecourse/contracts/, causing ENOENT right after a successful contracts generate
## Summary

`truecourse verify` fails with `ENOENT ... scandir '.../.truecourse/contracts'` **even though `contracts generate` just successfully wrote the contract files**, when the user picks the **"Stash and analyze committed state (recommended)"** option at the stash prompt.

Root cause: verify stashes the working tree with `git stash push --include-untracked` *before* reading contracts, and `.truecourse/contracts/` is an **untracked, not-ignored** directory — so the stash sweeps away the very contracts the verifier then tries to load. The pre-stash existence check passes and gives a false all-clear, making the error look contradictory ("generate wrote the files, but verify says they don't exist").

## Environment

- CLI: `truecourse@next` (installed via `npx truecourse@next`)
- Target repo under test: **directus/directus** — https://github.com/directus/directus, checked out at `/Users/.../TrueCourseTest/directus`
- Repo git state at time of `verify`: **0 modified, 259 untracked** files
- `.truecourse/` was **not committed** in the target repo (so its contents are untracked)
- OS: macOS

## Steps to reproduce

Reproduced against the public **directus/directus** repo (https://github.com/directus/directus). The trigger condition is general: any repo where `.truecourse/` is not committed (so the generated `contracts/` is untracked) and the working tree is dirty enough to prompt for stashing — directus is just a convenient, large public repo to try it on.

```bash
# 0. Clone the repo this was reproduced against (directus/directus)
git clone https://github.com/directus/directus
cd directus

# 1. Scan — produced 0 open conflicts
npx truecourse@next spec scan
#   → docs 48 · claims 117 · resolved 112 · decided 2 · open 0
#   → "No open conflicts — run `truecourse contracts generate`."

# 2. Generate — successfully wrote contract files
npx truecourse@next contracts generate
#   → "Wrote 2 contract files."
#   → .truecourse/contracts/unenforceable/entity.flow.undefined.tc
#   → .truecourse/contracts/unenforceable/auth.role.admin.unspecified.tc
#   → "Run `truecourse verify` to check code against the new contracts."

# 3. Verify — pick "Stash and analyze committed state (recommended)"
npx truecourse@next verify
```

### Actual result

```
│  ▲  Your repository has 0 modified and 259 untracked file(s).
│  ◇  How should TrueCourse handle them?
│     Stash and analyze committed state (recommended)
┌  Verify
  ✕ Loading contracts — ENOENT: no such file or directory, scandir '/Users/.../TrueCourseTest/directus/.truecourse/contracts'
  ○ Scanning code for operations
  ○ Comparing code against contracts
└  Failed: ENOENT: no such file or directory, scandir '/Users/.../TrueCourseTest/directus/.truecourse/contracts'
```

### Expected result

`verify` should read the contracts that `contracts generate` just wrote and report drifts (or "No drift detected"), regardless of the stash choice.

### 1. Ordering: existence check runs *before* the stash

In `packages/core/src/commands/spec-in-process.ts`, `verifyInProcess`:

- **L824** — `if (!fs.existsSync(contractsDir)) { throw ... "Run contracts generate first" }` — passes, because the contracts are on disk at this moment ✅
- **L844** — `result = await runWithStash(repoRoot, options.skipStash ?? false, tracker, () => verify({ contractsDir, codeDir }))`

So the order is: **check contracts exist → stash → read contracts**. The stash happens *between* the check and the read.

### 2. The stash sweeps away `.truecourse/contracts/`

`runWithStash` (same file, ~L972) stashes with `--include-untracked`:

```js
const res = await stashGit.stash(['push', '--include-untracked', '-m', 'truecourse-verify-stash']);
```

`--include-untracked` stashes **all untracked (but not ignored) files**. `.truecourse/contracts/` is untracked in the target repo, so it gets swept into the stash. Then `verify()` runs `scandir('.truecourse/contracts')` → **ENOENT**. The `finally` block does `git stash pop` afterward, so the contracts reappear on disk — but verify has already crashed.

### 3. Why `contracts/` is untracked-and-not-ignored (the deeper cause)

The `.truecourse/.gitignore` that TrueCourse scaffolds does **not** list `contracts/`.

`packages/core/src/config/paths.ts:11`:

```js
const GITIGNORE_CONTENTS = 'analyses/\nhistory.json\ndiff.json\nui-state.json\nlogs/\n.analyze.lock\n';
```

→ `contracts/`, `specs/`, and `verifier/` are missing from this list. Because they're not ignored, `git stash --include-untracked` grabs them. (`analyses/`, `logs/`, etc. *are* ignored, so the stash correctly leaves those alone — only the directories verify actually needs are unprotected.)

This also contradicts the README, which states contracts are **"gitignored by default"** — the code never actually ignores them.

## Summary of the defect

```
verify's stash (git stash push --include-untracked) removes the
untracked .truecourse/contracts/ directory, while the contracts
existence check runs BEFORE the stash and gives a false all-clear.

Contributing factor: .truecourse/.gitignore (paths.ts:11) does not
ignore contracts/ or specs/, contradicting the README's claim that
contracts are "gitignored by default".
```

## Workaround

Run verify without stashing:

```bash
npx truecourse@next verify --no-stash
```

`--no-stash` skips the stash entirely, so the contracts stay on disk. With 0 modified files, there's nothing in tracked code to stash anyway, so the `--no-stash` result is correct here.

## Proposed fix

Both parts are needed:

- **(A)** In `verifyInProcess` / `runWithStash`, make the stash exclude TrueCourse's own working dir (`.truecourse/`) so verify's inputs survive the stash, and/or move the contracts existence check to *after* the stash so the error is at least honest. The intent of the stash is to verify the committed **code** state — it should not be sweeping away TrueCourse-generated artifacts the verifier depends on.
- **(B)** Add `contracts/` and `specs/` to `GITIGNORE_CONTENTS` in `packages/core/src/config/paths.ts` so they're ignored by default (matching the README, and making `--include-untracked` skip them).

## References

- Reproduction repo: **directus/directus** — https://github.com/directus/directus
- `packages/core/src/commands/spec-in-process.ts` — `verifyInProcess` (existence check at L824, `runWithStash` call at L844, stash command at ~L972)
- `packages/core/src/config/paths.ts:11` — `GITIGNORE_CONTENTS` (missing `contracts/`, `specs/`)

DoilLee2222:
## What & why

Fixes #542. `truecourse verify` failed with `ENOENT ... scandir '.../.truecourse/contracts'` **immediately after a successful `contracts generate`**, whenever the user picked "Stash and analyze committed state" at the stash prompt.

### Root cause

`verify` (and `analyze`) neutralize uncommitted **code** by running `git stash push --include-untracked` so the run reflects committed state. But `--include-untracked` sweeps up *every* untracked file — including TrueCourse's own `.truecourse/` working dir. `verify` reads `.truecourse/contracts/` **during** the stash window, so the contracts it had just generated were gone → ENOENT. The pre-stash existence check passed first, which made the failure look contradictory ("generate wrote the files, but verify says they're missing").

The stash logic was also **duplicated** in two near-identical copies (`analyze-core.ts` and `spec-in-process.ts`), both with the same latent bug. `analyze` didn't visibly crash only because its inputs are source code read during the window, while its `.truecourse/` outputs are written after the pop.

## Changes

- **Extract one shared `runWithStash` helper** into `packages/core/src/lib/git.ts`, used by both `analyze` and `verify` (removes the duplicate copies). It scopes the stash to code only:
  ```
  git stash push --include-untracked -m <msg> -- . :(exclude,top).truecourse
  ```
  so TrueCourse's own working dir (contracts, specs, baselines) is never swept away. Uses the exported `TRUECOURSE_DIR` constant — no hardcoded `.truecourse` string.
- **Harden the subdirectory guard to compare realpaths.** `git --show-toplevel` returns the realpath, so a symlinked repo path (macOS `/tmp → /private/tmp`, a symlinked `$HOME`) made the old lexical `path.resolve` comparison mismatch and **silently skip the stash entirely** — analyze/verify then ran against the dirty tree while believing they were on committed state. Now `fs.realpathSync(repoRoot) === fs.realpathSync(gitRoot)`.
- **Repair the stale `.truecourse/.gitignore` template** (`config/paths.ts`) to match the documented committable-vs-local layout: ignore `contracts/`, `verifier/runs/`, `verifier/history.json`, `verifier/diff.json`, `.cache/`; keep `specs/`, `LATEST.json`, `verifier/LATEST.json`, `config.json` committable. This is defense-in-depth + makes the README's "gitignored by default" claim true. (New repos only; existing repos are protected by the stash fix regardless.)

## Tests

- `tests/core/run-with-stash.test.ts` (new) — proves the helper stashes dirty tracked code (run sees committed state) **and** leaves `.truecourse/contracts/` on disk; plus `skipStash`, clean-tree, and non-git cases.
- `tests/core/verify-store.test.ts` — added a `verifyInProcess` regression with contracts under the repo's own untracked `.truecourse/contracts/` (the #542 repro): verify reads them with no ENOENT.
- `tests/core/paths.test.ts` (new) — asserts the gitignore template ignores generated/local state but keeps the spec and baselines committable.

Full suite: **133 files / 4191 tests passing**, `tsc --noEmit` clean.

## Manual verification

Reproduce the original report against the public **directus/directus** repo (https://github.com/directus/directus):

```bash
git clone https://github.com/directus/directus && cd directus
truecourse spec scan
truecourse contracts generate
truecourse verify   # choose "Stash and analyze committed state"
```

Before: `ENOENT ... scandir '.../.truecourse/contracts'`. After: verify reads the contracts and reports drifts (or "No drift detected"). `truecourse analyze` on a dirty tree is unchanged.

### DIFF
--- a/apps/dashboard/server/package.json
+++ b/apps/dashboard/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.6",
+  "version": "0.6.6-next.0",
   "private": true,
   "type": "module",
   "scripts": {
--- a/apps/dashboard/server/package.json
+++ b/apps/dashboard/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.6-next.0",
+  "version": "0.6.7",
   "private": true,
   "type": "module",
   "scripts": {
--- a/packages/core/src/config/paths.ts
+++ b/packages/core/src/config/paths.ts
@@ -2,13 +2,34 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 
-const TRUECOURSE_DIR = '.truecourse';
+// Canonical name of the per-repo TrueCourse working directory. Exported as the
+// single source of truth so callers (e.g. the git stash helper, which must
+// never sweep this dir) reference the constant instead of a literal string.
+export const TRUECOURSE_DIR = '.truecourse';
+// Default `.truecourse/.gitignore`, mirroring the documented committable-vs-local
+// layout (see README "Setup"). Local-only state is ignored; the canonical spec
+// (`specs/`), the diff baselines (`LATEST.json`, `verifier/LATEST.json`) and
+// `config.json` stay committable so they travel with the repo.
+//
 // `LATEST.json` is intentionally tracked: it travels with the repo via git so
 // fresh clones and `git worktree add` checkouts inherit a baseline without
 // having to cold-start `truecourse analyze`. Convention: only commit it after
 // merging to main (post-merge analyze). Branch-local analyzes shouldn't
 // commit it, to avoid PR conflicts on a generated JSON.
-const GITIGNORE_CONTENTS = 'analyses/\nhistory.json\ndiff.json\nui-state.json\nlogs/\n.analyze.lock\n';
+const GITIGNORE_CONTENTS = [
+  'analyses/',
+  'history.json',
+  'diff.json',
+  'ui-state.json',
+  'logs/',
+  '.analyze.lock',
+  'contracts/',
+  'verifier/runs/',
+  'verifier/history.json',
+  'verifier/diff.json',
+  '.cache/',
+  '',
+].join('\n');
 
 // ---------------------------------------------------------------------------
 // Global paths (user-level)
--- a/packages/core/src/lib/git.ts
+++ b/packages/core/src/lib/git.ts
@@ -7,8 +7,10 @@
  * Also provides `isGitRepo()` for cases where callers want to check without throwing.
  */
 
+import fs from 'node:fs';
 import { simpleGit, type SimpleGit } from 'simple-git';
 import { createAppError } from './errors.js';
+import { TRUECOURSE_DIR } from '../config/paths.js';
 
 /**
  * Single-sourced message for the "this isn't a git repo" guard. TrueCourse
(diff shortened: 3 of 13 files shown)


apps/dashboard/server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.6",
+  "version": "0.6.6-next.0",
```
mushgev: Remember not to merge with this to main, you need to bump version to 0.6.7 when merging

packages/core/src/config/paths.ts
```diff
@@ -2,13 +2,34 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 
-const TRUECOURSE_DIR = '.truecourse';
+// Canonical name of the per-repo TrueCourse working directory. Exported as the
```
mushgev, line 5: 👍 on promoting this to an export, and `paths.ts` is the right home for it. But this string already had three identical local copies before this PR:

- `packages/core/src/config/paths.ts:5` (this one, now exported)
- `packages/core/src/lib/analysis-store.ts:25`
- `packages/core/src/lib/verify-store.ts:29`

This PR establishes `paths.ts` as the canonical exported source but only consumes it from the new `git.ts`, leaving the other two as their own hardcoded copies — so we end up with 1 canonical export + 2 stragglers, the same duplication this PR otherwise kills (it consolidated the duplicated stash logic for exactly this reason).

Could we also point `analysis-store.ts` and `verify-store.ts` at the new export in this PR? It's small and in-scope. If you'd rather defer it, fine — but let's make that a conscious call rather than leaving the constant half-deduped.

packages/core/src/lib/git.ts
```diff
@@ -43,3 +45,103 @@ export async function getGit(repoPath: string): Promise<SimpleGit> {
...
+        const gitRoot = (await stashGit.revparse(['--show-toplevel'])).trim();
+        // Skip stashing when the repo path is a subdirectory of a larger repo
+        // (e.g. test fixtures inside the main repo) — stashing there would
+        // touch unrelated parent-repo files. Compare *canonical* paths:
+        // `git --show-toplevel` returns the realpath, so a symlinked repoRoot
+        // (e.g. macOS /tmp → /private/tmp, or a symlinked $HOME) must be
+        // resolved too — otherwise we'd wrongly treat the real git root as a
+        // parent and silently skip the stash. When we do stash, simple-git's
+        // baseDir (repoRoot) makes the `.` pathspec resolve to the repo root.
+        if (fs.realpathSync(repoRoot) === fs.realpathSync(gitRoot)) {
```
mushgev, line 112: The realpath hardening is the right fix, but I don't think it's covered *deterministically*. Worth pinning down before merge.

Quirk: on macOS `os.tmpdir()` is itself symlinked (`/var → /private/var`), so `fs.mkdtempSync(os.tmpdir()+...)` in `run-with-stash.test.ts` already hands `runWithStash` a symlinked `repoRoot`, while `git --show-toplevel` returns the realpath. That means the first test ("stashes dirty tracked code") *does* exercise this branch on macOS — and would fail under the old lexical `path.resolve` comparison (stash silently skipped → `codeDuringRun` would be the dirty `x = 2`).

The catch: on Linux CI the tmpdir generally isn't symlinked, so there the comparison is true with **or** without this fix. So the regression we actually shipped (#542-adjacent: symlinked `$HOME`/`/tmp` silently skipping the stash and running against the dirty tree) has no deterministic guard — it'd pass green on CI even if someone reverted `realpathSync` back to `path.resolve`.

Could we add an explicit test that symlinks `repoRoot` regardless of platform — create a real dir, `fs.symlinkSync` a second path to it, pass the symlink as `repoRoot`, assert committed code is visible inside the run (i.e. the stash *did* happen)? That pins the realpath branch on every platform instead of relying on macOS's incidental `/var` symlink.

apps/dashboard/server/package.json
DoilLee2222: I just bump version to 0.6.7

packages/core/src/lib/git.ts
DoilLee2222: Added a cross-platform test that symlinks repoRoot and asserts the stash still happens. Confirmed it fails under the old lexical compare, so it pins the realpathSync branch on every platform, not just macOS.

PR #588, opened 2026-06-12
fix(analyzer): release tree-sitter WASM trees via shared withParsedTree helper

### LINKED ISSUE #584: Analyzer pipeline leaks tree-sitter WASM Trees → parser `Aborted()` in long-running dashboard → silently wrong analyze/verify results
## Summary

In a **long-running process** (the dashboard server), the tree-sitter **WASM runtime crashes (`Aborted()`)** because the **analyzer pipeline never frees the `Tree` objects it parses**. web-tree-sitter `Tree`s are backed by a fixed WASM-heap allocation that JS GC cannot reclaim — they must be released with an explicit `tree.delete()`. The analyzer leaks one tree per parsed file; over a long dashboard session (analyze + file-watcher re-runs across several large repos) the heap fills and the WASM module calls `abort()`.

Once the WASM module has aborted, **every subsequent `parser.parse()` fails for the rest of the process**. These failures are swallowed per-file as "non-fatal", so `analyze`/`verify` still return *successfully* — just with **empty code-side extraction**. The result is **silently wrong output**, not a crash.

This is the same bug class already fixed for the `contract-verifier` engine in **#532 / #533**, but the matching fix for the **analyzer** pipeline was never merged to `main` (it lives only on the stale `origin/fp-fixes` branch — commit `f2f4842a`).

## How it surfaced

Running `truecourse verify` on the same commit gave **different drift counts** depending on where it ran:

| Run | Process | Artifacts | `extractedOps` | Drifts |
|-----|---------|-----------|----------------|--------|
| `truecourse verify` (CLI) | fresh process | 67 | **1** | **28** (correct) |
| Dashboard "Verify" button | long-running server | 67 | **0** | **37** (wrong) |

The CLI and dashboard call the **identical** `verifyInProcess()` → `verify()` — same code, no logic difference. The divergence is purely process state: the CLI starts a clean process every time; the dashboard is one long-lived process whose WASM parser had already aborted.

Diffing the two saved run snapshots, the entire difference is in the `NamedConstant` comparator: in the dashboard run the Python code was never parsed (`extractedOps: 0`), so every spec constant was reported as `…​.no-code-counterpart` (info) instead of being matched. That inflated the drift count 28 → 37.

## Evidence

- Dashboard error log contained **6,731× `Aborted()`** (the WASM runtime, dead).
- The wrong (dashboard) snapshot has `extractedOperationCount: 0` — i.e. **no source file was parsed at all**.
- Running `verify()` **6× in a single fresh process** is perfectly stable at 28 → the engine itself is deterministic; the problem is accumulated process state, not the algorithm.
- The affected repo's working tree had **no modified tracked files** → rules out stash / working-tree-vs-committed as the cause.

## Root cause (two defects, compounding)

1. **Tree leak in the analyzer pipeline (primary).** Confirmed parse sites in `main` that never call `tree.delete()`:
   - `packages/analyzer/src/file-analyzer.ts` — `analyzeFile` (~L42) and `analyzeFileContent` (~L107)
   - `packages/core/src/services/violation-pipeline.service.ts` (~L404, per-file rule pass)
   - `packages/analyzer/src/extractors/entities.ts` (~L128)
   - `packages/analyzer/src/database-detector.ts` (~L117)
   - (any other direct `getParser().parse()` / `parseFile()` site in the analyzer should be audited as part of this)

   `tree.delete()` currently appears **only** in `packages/contract-verifier/*` — never in `packages/analyzer` or `packages/core`.

2. **Failures are silent (secondary / defense-in-depth).** `eachParsedSource` (`contract-verifier/src/extractor/source-walker.ts`) and the analyzer treat a per-file parse failure as non-fatal and skip it. A single bad file being skipped is reasonable, but a **dead WASM runtime makes *every* file fail** — which then looks identical to "no code found" and produces confidently-wrong results instead of an error.

## Why the verifier is already safe but the analyzer isn't

The disposal fix was applied to the verify/infer engine and merged, but the analyzer half was stranded on an unmerged branch:

| Fix | Commit | Scope | In `main`? |
|-----|--------|-------|-----------|
| "defer tree disposal to end of verify()/infer()" (#532) | `dd2f9c1c` | contract-verifier | ✅ |
| "eagerify handler facts, dispose trees per file" (#533) | `d4896c29` | contract-verifier | ✅ |
| **"fix(analyzer): release web-tree-sitter Tree handles after each parse"** | `f2f4842a` | **analyzer** (`file-analyzer.ts` + `violation-pipeline.service.ts`) | ❌ **only on `origin/fp-fixes`** |
| "perf(type-query): keep at most one scoped ts.Program alive" | `686463cc` | analyzer (TS-compiler memory) | ❌ only on `origin/fp-fixes` |

Because the dashboard runs `analyze` and `verify` in the **same process** sharing **one** WASM runtime, the analyzer's leak crashes the parser that the (already-fixed) verifier then depends on.

## Proposed fix — make disposal a shared primitive

Rather than re-patch each call site individually (which is how `f2f4842a` left the next site free to regress), factor tree-lifetime management into **one shared helper** in `@truecourse/analyzer` and route every parse site through it:

```ts
// e.g. packages/analyzer/src/parser.ts
export function withParsedTree<T>(
  filePath: string, code: string, lang: SupportedLanguage,
  fn: (tree: Tree) => T,   // MUST return plain data — no SyntaxNode may escape
): T {
  const tree = parseFile(filePath, code, lang);
  try { return fn(tree); } finally { tree.delete(); }
}
```

Then convert: `file-analyzer.ts` (both functions), `violation-pipeline.service.ts`, `extractors/entities.ts`, `database-detector.ts`, and any other analyzer parse site. Seed the disposal logic from `f2f4842a` (already tested), but expose it as the shared primitive both `analyze` and `verify` can use, instead of a per-file copy.

> ⚠️ Important constraint learned from #533: some extractors retain `SyntaxNode` references into the tree (e.g. handler bodies consumed downstream). Disposing the tree while such references are live caused `RuntimeError: memory access out of bounds`. Each converted site must eagerly reduce the tree to plain data before the `finally` runs.

Optionally also port `686463cc` (ts.Program LRU) since it addresses the related OOM during the type-aware rule pass.

### Secondary hardening
Make a dead WASM runtime fail loudly rather than silently. Options: detect repeated total parse failure and throw; or surface an `Aborted()`/parser-dead signal so `analyze`/`verify` report an error instead of returning empty extraction as a valid result.

## Acceptance criteria

- [ ] No analyzer parse site leaks — `tree.delete()` (via the shared helper) covers every `parseFile`/`getParser().parse()` in `packages/analyzer` and `packages/core`.
- [ ] A long-running dashboard process can run `analyze` + `verify` repeatedly across large repos with **zero `Aborted()`** in the logs.
- [ ] CLI `verify` and dashboard `verify` produce **identical** drift counts on the same commit.
- [ ] A dead/failed parser surfaces as an error, not a silently-empty result.
- [ ] Existing tests stay green (watch for `memory access out of bounds` regressions per the #533 constraint).

## References
- Prior art (merged): #532 (`dd2f9c1c`), #533 (`d4896c29`)
- Prior art (unmerged, analyzer): `f2f4842a`, `686463cc` on `origin/fp-fixes`

DoilLee2222:
Closes #584.

## Problem

web-tree-sitter `Tree` objects are backed by a fixed WASM-heap allocation that JS GC never reclaims — they must be released with an explicit `tree.delete()`. The **analyzer** pipeline parsed at four sites and freed at **none**, so a long-running process (the dashboard server, which runs `analyze` + the file watcher in-process across many repos) leaked one tree per parsed file until the WASM runtime exhausted its heap and called `abort()`.

Once the WASM module aborts, every subsequent parse fails — and `eachParsedSource` swallows those per-file failures as "non-fatal". So `verify` completes *successfully* with **empty code-side extraction**, silently inflating drift counts. That's the `28` (CLI, fresh process) vs `37` (dashboard, long-lived process) the user saw on aider: same commit, same engine, but the dashboard's WASM had already died.

The `contract-verifier` was already fixed for this class of bug in #532 / #533; the matching **analyzer** fix was authored on the unmerged `fp-fixes` branch (`f2f4842a`) and never landed in `main`.

## Fix — one shared primitive

A single `withParsedTree(filePath, code, language, use)` helper in `@truecourse/analyzer` that parses, runs `use`, and disposes the tree in a `finally`. Every analyzer parse site now routes through it instead of holding a `Tree` by hand:

| File | Change |
|------|--------|
| `analyzer/src/parser.ts` | new `withParsedTree` helper |
| `analyzer/src/index.ts` | export it |
| `analyzer/src/file-analyzer.ts` | `analyzeFile` + `analyzeFileContent` via the helper; extraction deduplicated into one `buildFileAnalysis` (also fixes a latent gap — `analyzeFileContent` never handled Python) |
| `analyzer/src/extractors/entities.ts` | `detectEntities` via the helper; dead lazy `parser` getter + unused imports removed |
| `core/src/services/violation-pipeline.service.ts` | per-file rule pass via the helper; dropped a no-op `changedFileSet ? filePath : filePath` ternary |

`use` must reduce the tree to plain data before returning — no `SyntaxNode` escapes (the `memory access out of bounds` trap from #531). All converted sites already return plain data (`FileAnalysis`, `Entity[]`, violation objects).

**Intentionally untouched:** the language extractors' `getParser()` calls only read `parser.language` to build a `Query` (no parse, no leak), and `database-detector`'s parser is a regex-based schema parser — neither leaks. `contract-verifier` already disposes its own trees (#532/#533).

## Verification

End-to-end on aider (147 Python files) — a long-lived process doing **8,800 analyze passes** then `verify`, vs a fresh-process CLI baseline:

| Scenario | `extractedOps` | verify drifts | RSS |
|----------|:---:|:---:|-----|
| CLI verify (fresh) | 1 | **29** | — |
| Dashboard sim — **fixed** | 1 | **29** ✅ | 366→632 MB (flat) |
| Dashboard sim — old/leaky | **0** | **40** ❌ | 365→2082 MB → WASM aborts @ round 66 |

The old code reproduces the exact failure (WASM abort → `extractedOps=0` → drift inflated to all-`info` `no-code-counterpart`); the fix keeps CLI and dashboard in agreement.

- ✅ `tsc` clean (`@truecourse/analyzer`, `@truecourse/core`)
- ✅ Full suite: **4246 tests pass** (136 files)
- ✅ 40k-parse stress: bounded RSS, zero `Aborted()`
- Net **−12 lines**

### DIFF
--- a/apps/dashboard/server/package.json
+++ b/apps/dashboard/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.7",
+  "version": "0.6.7-next.1",
   "private": true,
   "type": "module",
   "scripts": {
--- a/apps/dashboard/server/package.json
+++ b/apps/dashboard/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.7-next.1",
+  "version": "0.6.8",
   "private": true,
   "type": "module",
   "scripts": {
--- a/packages/analyzer/src/extractors/entities.ts
+++ b/packages/analyzer/src/extractors/entities.ts
@@ -6,8 +6,8 @@
 import type { FileAnalysis, Entity, EntityField, EntityRelationship, SupportedLanguage } from '@truecourse/shared'
 import { dataLayerPatterns } from '../patterns/layer-patterns.js'
 import { matchesPattern } from '../patterns/index.js'
-import type { Node as SyntaxNode, Tree } from 'web-tree-sitter'
-import { getParser } from '../parser.js'
+import type { Node as SyntaxNode } from 'web-tree-sitter'
+import { withParsedTree } from '../parser.js'
 
 /**
  * Check if a file likely contains entities based on ORM import patterns
@@ -101,12 +101,6 @@ function mapToSimpleType(type: string): string {
 }
 
 class TypeScriptEntityDetector {
-  // Lazy: do not call getParser() at construction time — callers may
-  // instantiate this before initParsers() has completed.
-  private get parser() {
-    return getParser('typescript')
-  }
-
   shouldScanFile(filePath: string): boolean {
     if (!/\.(ts|tsx|js|jsx)$/.test(filePath)) {
       return false
@@ -125,32 +119,29 @@ class TypeScriptEntityDetector {
   }
 
   detectEntities(sourceCode: string, filePath: string, service: string): Entity[] {
-    const tree = this.parser.parse(sourceCode)
-    const entities: Entity[] = []
+    return withParsedTree(filePath, sourceCode, 'typescript', (tree) => {
+      const entities: Entity[] = []
 
-    if (!tree) return entities
-
-    // Find all class declarations
-    const classNodes = this.findClassDeclarations(tree.rootNode)
-
-    for (const classNode of classNodes) {
-      const entity = this.detectEntityFromClass(classNode, sourceCode, filePath, service)
-      if (entity) {
-        entities.push(entity)
+      // Find all class declarations
+      const classNodes = this.findClassDeclarations(tree.rootNode)
+      for (const classNode of classNodes) {
+        const entity = this.detectEntityFromClass(classNode, sourceCode, filePath, service)
+        if (entity) {
+          entities.push(entity)
+        }
       }
-    }
 
-    // Also find interface declarations (for projects using plain interfaces as models)
-    const interfaceNodes = this.findInterfaceDeclarations(tree.rootNode)
-
-    for (const interfaceNode of interfaceNodes) {
-      const entity = this.detectEntityFromInterface(interfaceNode, sourceCode, filePath, service)
-      if (entity) {
-        entities.push(entity)
+      // Also find interface declarations (for projects using plain interfaces as models)
+      const interfaceNodes = this.findInterfaceDeclarations(tree.rootNode)
+      for (const interfaceNode of interfaceNodes) {
+        const entity = this.detectEntityFromInterface(interfaceNode, sourceCode, filePath, service)
+        if (entity) {
+          entities.push(entity)
+        }
       }
-    }
 
-    return entities
+      return entities
+    })
   }
 
   private findClassDeclarations(node: SyntaxNode): SyntaxNode[] {
--- a/packages/analyzer/src/file-analyzer.ts
+++ b/packages/analyzer/src/file-analyzer.ts
@@ -1,7 +1,7 @@
 import { readFile } from 'fs/promises'
 import type { FileAnalysis, SupportedLanguage, ExportStatement } from '@truecourse/shared'
 import { detectLanguage } from './language-config.js'
-import { parseFile } from './parser.js'
+import { withParsedTree, type Tree } from './parser.js'
 import { extractCalls, buildFunctionContext } from './extractors/calls.js'
 import { extractHttpCalls } from './extractors/http-calls.js'
 import { extractRouteRegistrations } from './extractors/route-registrations.js'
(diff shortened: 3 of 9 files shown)


apps/dashboard/server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@truecourse/dashboard-server",
-  "version": "0.6.7",
+  "version": "0.6.7-next.1",
```
mushgev: make sure to update the version before merging

DoilLee2222: will bump version to 0.6.8

PR #607, opened 2026-06-20
chore(fp): start discovery for abpframework/abp

DoilLee2222:
## fp-discover: starting discovery for `abpframework/abp`

This PR marks the `abpframework/abp` campaign as `status: discovering` in `docs/fp-automation/campaigns.yaml`, kicking off an **fp-discover** run under the `cs-` scope (`TECH_STACKS=csharp`).

### What happens next
- Building truecourse from local source (`pnpm build:dist`) and analyzing `abpframework/abp` with `analyze --no-llm --no-stash --no-skills`.
- Triaging violations by rule, classifying TP / FP / borderline.
- Filing one `[cs-fp-fix]` issue per rule with a ≥10% FP rate.
- Updating this campaign's `baseline.*` block and pushing the result to this same PR.

This PR is labelled `cs-fp-discover` so that merging it fires the **fp-next-fix** routine and starts the inner fix loop. Merge it when you're ready to start that loop — discovery work continues on this PR regardless.

cc @mushgev

---
_Generated by [Claude Code](https://claude.ai/code/session_01CWSuGmT1SeykPGxQyZbbSz)_


DoilLee2222:
### Borderline rules (no issue filed — human triage)

These C# rules fire heavily but their detections are *accurate* (or the underlying advice is defensible); the question is policy/applicability, not a detection bug. Per the routine they get no FP issue — flagging here for a human call. Add `fp-confirmed` / `tp-confirmed` as you see fit.

- **`style/deterministic/docstring-completeness`** — 12,752 on `.cs` (59% of all C# violations). Each report ("public class X has no XML doc comment") is literally true, but it flags *every* undocumented public type/member. On a framework of this size that is overwhelming noise (marker classes, modules, resources, DTOs). Decision needed: should this rule require docs on every public symbol, be opt-in, or skip framework/marker classes?
- **`code-quality/deterministic/magic-number`** — 1,981 on `.cs`. Valid detections, but noisy on self-explanatory constants (HTTP status boundaries `200`/`300`, stream buffer size `4096`, `AddYears(10)`).
- **`code-quality/deterministic/too-many-positional-arguments`** — 348 on `.cs`. Mostly DI constructors (e.g. 10 injected services). The remediation "group related values into a record or options class" does not apply to constructor-injected dependencies; the count is accurate but the advice is often inapplicable in C#.
- **`code-quality/deterministic/parameter-reassignment`** — 242 on `.cs`. Many are the idiomatic optional-parameter pattern `param ??= default;`, which most C# code treats as normal.
- **`bugs/deterministic/empty-catch`** — 50 on `.cs`. Several flagged catches already document an intentional ignore with a comment and catch a specific exception type — which is exactly what this rule's own remediation text recommends — yet they are still flagged.
- **`performance/deterministic/sync-fs-in-request-handler`** — 50 on `.cs`. The "blocks a thread-pool thread on the async path" advice is reasonable, but the "request handler" framing misapplies to CLI/console code (most hits are in `Volo.Abp.Cli.*`), where blocking I/O is not a scalability concern.

### Note on the JS/TS portion of this repo

68,970 total violations, but only **21,453 are in `.cs` files**. The remaining ~47k are in `.js`/`.ts` files — overwhelmingly the JS/TS rules firing (largely correctly) on **vendored libraries** under `wwwroot/libs/` (bootstrap, jQuery, datatables, …) and Razor page scripts. Those are JS rules on JS files, not C# FPs, so they are out of scope for this campaign. A separate, larger question for the team: should the analyzer skip vendored/third-party asset directories (`wwwroot/libs`, `*.min.js`) entirely? That alone would remove the bulk of this repo's violation count.

---

DoilLee2222:
### fp-discover summary — abpframework/abp

- **Target ref:** `38746937a5bb9ab51a0d74ac7df122cd3eb0b847`
- **Total violations:** 68,970 (21,453 in `.cs`; ~47k in vendored/Razor JS/TS)
- **Sampled C# baseline TP rate:** **0.72** (tp 92 / fp 35 over the triaged C# rules; borderline excluded)
- **FP issues filed:** 7

| # | rule_key | clear FPs in sample |
|---|---|---|
| #611 | `bugs/deterministic/ignored-return-value` | `StringBuilder.Append` treated as a no-op/lazy-LINQ result |
| #612 | `code-quality/deterministic/redundant-jump` | `return this;` (fluent, non-void) misread as redundant `return;` |
| #613 | `bugs/deterministic/self-assignment` | object-initializer `new T { Prop = Prop }` |
| #614 | `code-quality/deterministic/unnecessary-namespace-qualifier` | fires on `namespace A.B.C;` declarations |
| #615 | `bugs/deterministic/fstring-missing-placeholders` | escaped-brace `$"{{{expr}}}"` miscounted as no-hole |
| #616 | `code-quality/deterministic/no-extraneous-class` | type-token marker class can't be `static` |
| #617 | `style/deterministic/csharp-naming-convention` | EF Core migration classes (generated underscore names) |

Borderline / policy rules (docstring-completeness, magic-number, too-many-positional-arguments, parameter-reassignment, empty-catch, sync-fs-in-request-handler) are listed in the comment above for human triage.

Each issue carries `cs-fp-fix` + `fp-target:abpframework-abp`. The campaign stays `status: discovering`; baseline recorded in `campaigns.yaml` on this branch. Merging this PR (labelled `cs-fp-discover`) starts the fp-next-fix inner loop.

Note: this branch also carries a one-line build fix (`scripts/build.ts`) — a fresh-checkout `pnpm build:dist` was failing because `@truecourse/llm` was never built before `spec-consolidator`, which imports it. That blocked the discovery build, so it's fixed here.

---

PR #621, opened 2026-06-21
C# analyzer: 253 tree-sitter rules across all 8 domains

mushgev:
Implements the **build-free (tree-sitter) C# rule track** — 253 new deterministic rules, bringing C# from 308 → **561 analyzer rules**. This closes the syntactically-detectable rules; the type-dependent rules are the follow-up Roslyn-host track.

## By domain (new rules)
| domain | new |
|---|---|
| code-quality | 100 |
| bugs | 49 |
| security | 38 |
| performance | 30 |
| architecture | 16 |
| reliability | 11 |
| style | 8 |
| database | 1 |

## Each rule
Rule definition (`<domain>/deterministic.ts`) + tree-sitter visitor (`<domain>/visitors/csharp/`) + registration + positive/negative unit tests + an isolated negative-fixture marker — all false-positive-verified to zero against the fixture corpus.

## Tooling fixes
- `scripts/regenerate-expected-graph.ts` was missing `initParsers()` and only emitted the JS/Python graph schema — now inits parsers and emits `fileDependencies`/`routes`/`databases`.
- Fixed a systemic bug where tree-sitter exposes `//` comments as named children, causing block-statement-counting visitors to miscount.

## Validation
Full analyzer suite green throughout — **4405 tests, 50 files, zero JS/Python regressions**.

## Not in this PR
~210 rules need a semantic model (type resolution / dataflow / cross-file analysis) to detect without false positives — those are the Roslyn-host track (#622).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #622, opened 2026-06-21
feat(csharp): Roslyn semantic-host skeleton (Phase 1)

mushgev:
Starts the **C# semantic backend** — the Roslyn host from `docs/CSHARP_OUTPERFORM_SONAR_PLAN.md`. This is the gate to the ~210 rules deferred from the tree-sitter track (#621): everything needing resolved types, dataflow, nullability, or cross-file symbols.

## What this is
A small .NET worker the Node analyzer spawns as a child process (the same pattern as the Pyright LSP), speaking newline-delimited JSON over stdin/stdout. We run **our own** rules on Roslyn's semantic model and return them in our violation schema — clean-room, no third-party analyzer bundled.

## In this PR
- Host skeleton: stdio JSON protocol (`ping`/`analyze`), in-process `CSharpCompilation`, `ISemanticRule` registry.
- First real semantic rule `referenceequals-on-value-type` (Sonar S2995 / CA2013) — **validated end-to-end**: `ReferenceEquals(int, int)` is flagged (impossible for tree-sitter — needs to know `int` is a value type), `ReferenceEquals(object, object)` is not.
- Builds clean against .NET 8 (`dotnet build`, restores `Microsoft.CodeAnalysis.CSharp`).

## Next (tracked in the README)
Node client + `registry.ts` wiring (fall back to tree-sitter when the SDK is absent) → MSBuildWorkspace project loading → port the ~210 deferred rules as `ISemanticRule`s in fixture-first waves.


mushgev:
Consolidated into #625 (single C# PR). All commits from this branch are contained in #625; closing to avoid the stacked-PR split and the catalog merge conflicts it caused.

PR #624, opened 2026-06-21
feat(csharp): wire the Roslyn host into the analyze pipeline

mushgev:
Stacked on #622 (base = `csharp-roslyn-host`). Makes the C# semantic host **live in the product** — its rules flow through `truecourse analyze` end-to-end.

## What's wired
- **Node client** (`roslyn-host-client.ts`) — spawns the host as a batch child process, sends C# files + enabled host-rule keys, returns violations. **Fail-hard**: missing host/.NET runtime → `RoslynHostUnavailableError`, no tree-sitter fallback.
- **`engine` field** on `AnalysisRule` (`tree-sitter` default | `roslyn-host`). language-support derives host rules as `csharp: supported` (no visitor), `not-applicable` elsewhere.
- **Pipeline hook** — after the tree-sitter pass, one batch host call over all C# files; violations enriched into `CodeViolation` (severity/title from the rule, snippet from source).
- **Bug fix** — `toAnalysisRule` was dropping the new `engine` field, so host rules never reached the pipeline.

## Validation
- Client round-trip: 4 tests (value-type flagged, ref-type clean, allow-list, fail-hard).
- **End-to-end**: a value-type `ReferenceEquals` surfaces in the store from a real analyze (correct severity/title/path, once).
- Full analyzer + C# analyze suites green — **3889 tests**, no regressions (the existing C# analyze now invokes the host too).

## Next
Port the ~210 deferred rules as `ISemanticRule` walkers in the host (waves), then MSBuildWorkspace project loading.

### DIFF
--- a/packages/analyzer/src/index.ts
+++ b/packages/analyzer/src/index.ts
@@ -196,3 +196,11 @@ export async function analyzeRepository(rootPath: string): Promise<{
 
 // Config file checkers
 export { checkPyprojectToml } from './rules/bugs/pyproject-checker.js'
+
+export {
+  runRoslynHost,
+  resolveRoslynHostBinary,
+  RoslynHostUnavailableError,
+  type RoslynFile,
+  type RoslynHostViolation,
+} from './roslyn-host-client.js'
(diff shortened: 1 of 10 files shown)


mushgev:
Consolidated into #625 (single C# PR). All commits from this branch are contained in #625; closing to avoid the stacked-PR split and the catalog merge conflicts it caused.

Issue #679, opened 2026-07-01
[cs-fp-campaign-stuck] abpframework/abp


DoilLee2222:
## Campaign stuck — automation cannot self-progress

The `fp-next-fix` routine (`SCOPE=cs-`, `TECH_STACKS=csharp`) ran with an
**empty pickable queue** and **0 successes**. Every open `cs-fp-fix` issue for
the `abpframework/abp` campaign carries either `cs-fp-blocked` or a **stale**
`cs-fp-in-progress` lock, so the per-issue loop had no pickable work.

This is a fresh tracking issue — the previous one (#671) was closed after
PR #678 landed the Roslyn-host fixes. The state has since changed in ways that
need a human hand (details below).

### TP rate

| | Value |
|---|---|
| Current `tp_rate` (baseline `8665ce0a23622f658b531b0627c7c025935fdb3b`) | **0.83** (tp 357 / fp 71) — last measured value |
| Close threshold | **0.90** |

**Re-measurement was not possible in this routine's environment.** C# analysis
requires the Roslyn semantic host (`tools/csharp-roslyn-host`, .NET 8 SDK), and
`dotnet` is not on `PATH` here (`dotnet: command not found`). C# analysis
**fails hard** with no tree-sitter-only fallback, by design, so the queue-empty
re-analyze / TP-recompute could not run. This is the same environmental
limitation #671 recorded. The baseline **0.83** stands as the last known rate.

> Note: this environmental gap also means the `cs-` routine cannot *verify* any
> tree-sitter C# rule fix (steps 7/9 run `pnpm test`, which exercises the C#
> analyzer through the host, and the after-loop delta re-analyzes the target).
> Without .NET in the routine's environment, the cs- fp-fix loop cannot complete
> a fix end-to-end even for a tree-sitter-only rule. Worth fixing at the
> environment/design level.

### Action needed from a human (or a .NET-capable run)

**1. Verify + close the 6 issues PR #678 already fixed.** PR #678
("fundamentally fix all 6 blocked Roslyn C# false positives", merged
2026-07-01) rewrote the Roslyn-host rules for these, but its body used
`Closes #677` (the tracking issue) rather than closing each individual issue —
so they are still **open + `cs-fp-blocked`** and keep the queue jammed:

- [ ] #640 — `code-quality/deterministic/type-name-matches-namespace` — fixed in #678 (skip nested types; add own-enclosing-namespace pattern)
[...]

Issue #676, opened 2026-06-30
[cs-fp-fix] bugs/deterministic/check-against-value-being-assigned in abpframework/abp


DoilLee2222:
```yaml
target_repo: abpframework/abp
target_ref: 8665ce0a23622f658b531b0627c7c025935fdb3b
rule_key: bugs/deterministic/check-against-value-being-assigned
fp_count: 32
samples:
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/feature-management/src/Volo.Abp.FeatureManagement.Domain/Volo/Abp/FeatureManagement/FeatureDefinitionRecord.cs#L138''````
    why_fp: "`if (Name != otherRecord.Name) { Name = otherRecord.Name; }` where `Name` is a `public virtual { get; set; }` property on an EF Core aggregate root (`FeatureDefinitionRecord : BasicAggregateRoot<Guid>`). Assigning a tracked entity property marks it modified; the `!=` guard deliberately avoids dirtying the entity when unchanged, so the assignment is NOT a redundant no-op."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/feature-management/src/Volo.Abp.FeatureManagement.Domain/Volo/Abp/FeatureManagement/FeatureGroupDefinitionRecord.cs#L57''````
    why_fp: "Same entity `Patch`-style guarded property assignment on a change-tracked aggregate root property."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/permission-management/src/Volo.Abp.PermissionManagement.Domain/Volo/Abp/PermissionManagement/PermissionDefinitionRecord.cs#L140''````
    why_fp: "Guarded assignment to a virtual entity property; the guard prevents spurious change-tracking/audit writes."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/permission-management/src/Volo.Abp.PermissionManagement.Domain/Volo/Abp/PermissionManagement/PermissionGroupDefinitionRecord.cs#L56''````
    why_fp: "Guarded assignment to a virtual entity property on a change-tracked record."
[...]

Issue #675, opened 2026-06-30
[cs-fp-fix] code-quality/deterministic/property-name-matches-get-method in abpframework/abp


DoilLee2222:
```yaml
target_repo: abpframework/abp
target_ref: 8665ce0a23622f658b531b0627c7c025935fdb3b
rule_key: code-quality/deterministic/property-name-matches-get-method
fp_count: 29
samples:
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/WebClientInfo/HttpContextWebClientInfoProvider.cs#L27''````
    why_fp: "`public string? BrowserInfo => GetBrowserInfo();` — the property DELEGATES to the `GetBrowserInfo()` method; it is a single source of truth (the public surface over an overridable protected getter), not two competing ways to obtain the value."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore/Volo/Abp/AspNetCore/WebClientInfo/HttpContextWebClientInfoProvider.cs#L29''````
    why_fp: "`public string? ClientIpAddress => GetClientIpAddress();` — expression-bodied property delegating to its own `GetX()` method."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore.Components.Web.Theming/Theming/DefaultThemeManager.cs#L10''````
    why_fp: "`public ITheme CurrentTheme => GetCurrentTheme();` — property is implemented by calling the method; one implementation, not a confusing duplicate."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore.Mvc.Dapr/Volo/Abp/AspNetCore/Mvc/Dapr/DaprAppApiTokenValidator.cs#L13''````
    why_fp: "`protected HttpContext HttpContext => GetHttpContext();` — delegating expression-bodied property."
[...]

Issue #674, opened 2026-06-30
[cs-fp-fix] code-quality/deterministic/missing-access-modifier in abpframework/abp


DoilLee2222:
```yaml
target_repo: abpframework/abp
target_ref: 8665ce0a23622f658b531b0627c7c025935fdb3b
rule_key: code-quality/deterministic/missing-access-modifier
fp_count: 5
samples:
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.EntityFrameworkCore/Volo/Abp/Domain/Repositories/EntityFrameworkCore/EfCoreRepository.cs#L27''````
    why_fp: "`DbContext IEfCoreRepository<TEntity>.DbContext => ...` is an EXPLICIT interface implementation of a property. C# forbids an access modifier on explicit interface members — adding one is a compile error — so flagging it as 'missing access modifier' is impossible to act on."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.Localization/Volo/Abp/Localization/AbpDictionaryBasedStringLocalizer.cs#L303''````
    why_fp: "`LocalizedString IStringLocalizer.this[string name] => ...` is an explicit interface implementation of an INDEXER; explicit interface members take no access modifier."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.Localization/Volo/Abp/Localization/AbpDictionaryBasedStringLocalizer.cs#L305''````
    why_fp: "Explicit interface indexer implementation (`IStringLocalizer.this[string, params object[]]`) — no modifier allowed."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/MongoEntityModelBuilder.cs#L24''````
    why_fp: "`BsonClassMap IMongoEntityModelBuilder.BsonMap => ...` explicit interface property implementation — no modifier allowed."
  - url: ````''https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.MongoDB/Volo/Abp/MongoDB/MongoEntityModelBuilder.cs#L26''````
[...]

PR #682, opened 2026-07-01
test(cs-fp): add true-bug negative fixtures for #680's narrowed C# rules

DoilLee2222:
## Problem

[PR #680](https://github.com/truecourse-ai/truecourse/pull/680) narrows five C# rules to drop false positives, but ships only **positive** fixtures (assert zero violations). The [`fp-next-fix` routine](https://github.com/truecourse-ai/truecourse/blob/main/docs/fp-automation/prompts/fp-next-fix.md) requires **both** — step 5 adds the FP to the positive fixture, and **step 6** adds a paraphrased *true-bug* case to the negative fixture, proving the narrowed rule still catches the real bug. #680 relied on pre-existing negative markers instead of adding the discriminating near-miss that sits right next to each exempted FP shape.

This PR adds those missing negative fixtures.

## Fixtures added

All three are marker-driven (`// VIOLATION: <rule-key>`) files under `sample-csharp-project-negative`, each the true-bug counterpart of one of #680's positive fixtures:

| Negative fixture | Rule(s) that must still fire | Discriminates from the FP #680 exempted |
|---|---|---|
| `Bugs/GuardedLocalAssignmentBug.cs` | `bugs/deterministic/check-against-value-being-assigned` | #676 exempts a guarded write to an **enclosing-type property** (observable setter). A guard on a plain **local** is a genuine no-op → still fires. |
| `CodeQuality/NonDelegatingGetAccessors.cs` | `code-quality/deterministic/property-name-matches-get-method`, `code-quality/deterministic/property-matches-get-method` | #675 exempts a property whose getter **delegates** to `GetX()`. A stored property + a `GetX()` that computes a **different** result is the confusing colliding pair → still fires. |
| `CodeQuality/ImplicitPrivateMembersBug.cs` | `code-quality/deterministic/missing-access-modifier`, `code-quality/deterministic/unused-private-member` | #674 exempts **explicit interface** members (which forbid a modifier). An ordinary member with no modifier, and a genuinely unread private field, are real → still fire. |

The two incidental co-fires in `ImplicitPrivateMembersBug.cs` (`static-method-candidate` on the helper, and `docstring-completeness` resolved by giving `Total` a real doc comment) are handled so the strict Level-2 harness stays clean.

## Verification

Loose-text C# negative harness green — every fire in the new files sits on a marker, and every tree-sitter marker fires:

```
tests/analyzer/csharp-negative.test.ts  (6 tests) — 6 passed
```

`property-matches-get-method` is a `roslyn-host` rule, so its marker is verified in CI where the .NET 8 host is built (this environment has no .NET, mirroring #679). The shape is non-delegating, so it is not caught by #680's delegation exemption.

These true-bug shapes fire on the current `main` visitors as well as post-#680 — #680 only *narrowed* the rules, so the near-misses were never in the exempted set.

cc @mushgev

---
_Generated by [Claude Code](https://claude.ai/code/session_01NaE3Ze6azv18GrxtrYy5a9)_

### DIFF
--- a/tests/fixtures/sample-csharp-project-negative/expected-graph.json
+++ b/tests/fixtures/sample-csharp-project-negative/expected-graph.json
@@ -481,6 +481,12 @@
       "layer": "api"
     },
     {
+      "name": "GuardedLocalAssignmentBug",
+      "service": "ApiGateway",
+      "kind": "class",
+      "layer": "service"
+    },
+    {
       "name": "HandlerBase",
       "service": "ApiGateway",
       "kind": "class",
@@ -620,6 +626,12 @@
     },
     {
       "name": "IMetricsRecorder",
+      "service": "ApiGateway",
+      "kind": "class",
+      "layer": "service"
+    },
+    {
+      "name": "ImplicitPrivateMembersBug",
       "service": "ApiGateway",
       "kind": "class",
       "layer": "service"
@@ -896,6 +908,12 @@
     },
     {
       "name": "NetworkSection",
+      "service": "ApiGateway",
+      "kind": "class",
+      "layer": "service"
+    },
+    {
+      "name": "NonDelegatingGetAccessors",
       "service": "ApiGateway",
       "kind": "class",
       "layer": "service"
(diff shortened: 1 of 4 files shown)


DoilLee2222:
Superseded — the negative fixtures have been moved into #680 (which is still open and is where the rule fixes live), so they ship together in one PR. Closing this in favor of #680.

---

PR #721, opened 2026-07-06
docs(fp): add fp-next-fix-review routine prompt

DoilLee2222:
## What

Adds a new autonomous routine prompt, `docs/fp-automation/prompts/fp-next-fix-review.md`, that reviews an open **fp-fix batch PR** produced by `fp-next-fix` and takes exactly one terminal action:

- **Compliant** → **merges the PR** (which fires the next `fp-next-fix` session via Trigger B, keeping the inner loop running with no human in the loop).
- **Non-compliant** → **opens a `[<SCOPE>fp-review-failed]` issue** (`cc @mushgev`) and comments on the PR, leaving it open for a human. The issue is the notification signal.

`fp-next-fix.md` remains the single source of truth — this prompt points at its sections rather than restating rules.

## Review checklist coverage

The prompt verifies everything observable in the PR artifact:

- Branch prefix `claude/<SCOPE>fp-fix/batch-…` and `<SCOPE>fp-fix` label
- Title/commit format `fix(fp): resolve <N> FPs from <owner>/<repo>`, with `N` = `Closes` count, `1 ≤ N ≤ 5` (the 5-success cap)
- Changed files in-bounds (`rules/`, `patterns/`, `sample-*/` only)
- Fixture rules: no `// VIOLATION:` in positive, `// VIOLATION: <rule-key>` in negative, no OSS-project identity, no copy-pasted code
- Visitor/pattern changes are scoped, not cross-module refactors
- PR-body completeness (`Closes`, `## Fixes` table, per-rule sections, `cc @mushgev`) with internal count consistency
- Body-vs-diff cross-checks: fixtures named in the body actually appear in the diff, each rule adds **both** a positive and negative fixture plus a visitor edit
- `## FP-count delta` present (table or `unavailable:`) **and** direction sanity (a fixed rule's count must actually drop)
- CI green (waits if still running)
- Linked issues carry `<SCOPE>fp-fix` + `<SCOPE>fp-in-progress`
- No Claude-session details in commits/body

Scope-parameterized (`SCOPE` / `TECH_STACKS`) to match the sibling routines, with the standard commit/PR hygiene block.

## Notes

- Intended trigger: `pull_request.opened` / `reopened` on head branch prefix `claude/<SCOPE>fp-fix/batch-` — a distinct event from Trigger B (`closed`+merged), so the two don't collide.
- Docs-only change; no code or tests affected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01345gijHUZQQaC56rDnGxDT)_

### DIFF
--- a/docs/fp-automation/prompts/fp-next-fix-review.md
+++ b/docs/fp-automation/prompts/fp-next-fix-review.md
@@ -1,0 +1,266 @@
+# fp-next-fix-review routine prompt
+
+You are the **fp-next-fix-review** routine. You run inside an
+Anthropic-managed cloud session, autonomously, with no human in the
+loop. Your job is to **review a single open fp-fix PR** produced by the
+`fp-next-fix` routine, decide whether it faithfully follows every
+instruction in
+[`docs/fp-automation/prompts/fp-next-fix.md`](./fp-next-fix.md), and then
+take exactly one terminal action:
+
+- **Compliant** → **merge the PR**. The merge fires the next
+  `fp-next-fix` session (Trigger B), so the inner loop keeps running
+  with no human in the loop.
+- **Non-compliant** → **do not merge**. Open one issue on
+  `truecourse-ai/truecourse` describing every violation, and comment
+  the same summary on the PR. (The issue is the human signal; leave the
+  PR open for a human to fix or close.)
+
+`fp-next-fix.md` is the **single source of truth** for what "compliant"
+means. This prompt does not restate its rules — it points at them. When
+in doubt about whether the PR satisfies a requirement, re-read the
+relevant section of `fp-next-fix.md` and judge against it, not against
+your memory.
+
+## Routine parameters (scope)
+
+This prompt is **scope-parameterized** so more than one account can run the same chain over
+disjoint campaign sets without colliding. The invoking routine prompt (the bootstrap pointer)
+supplies two values; treat either as empty when omitted — the default account's behavior,
+byte-identical to an unscoped run.
+
+- **`SCOPE`** — a prefix applied to **every** branch, issue label, and issue-title tag this routine
+  creates **or** searches. Wherever this document shows `<SCOPE>`, substitute it verbatim. Default
+  **empty** → `claude/fp-fix/…`, label `fp-fix`, title `[fp-…]`. The C# account
+  uses `SCOPE=cs-` → `claude/cs-fp-fix/…`, label `cs-fp-fix`, title `[cs-fp-…]`.
+  **Never touch another scope's tokens** — the branch prefix is the unique trigger (labels are not
+  trigger filters), so the prefix is what isolates the accounts.
+- **`TECH_STACKS`** — a comma-separated allow-list of campaign tech stacks this routine may act on,
+  matched against each campaign's `tech_stack` in `campaigns.yaml`. Default **empty = no filter**;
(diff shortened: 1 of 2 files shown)


docs/fp-automation/prompts/fp-next-fix-review.md
```diff
@@ -0,0 +1,266 @@
...
+routine.) Any changed file outside the allow-list → violation, listing
+the offending path.
+
+### 4. Fixture rules
+
+For each added/changed fixture file:
+- Files under a `*-positive` fixture must contain **no** `// VIOLATION:`
+  comment (fp-next-fix step 5).
+- Files under a `*-negative` fixture must mark the offending line with
+  `// VIOLATION: <rule-key>` (or the Python-appropriate comment)
```
mushgev, line 116: Can you check why there is separate logic for Python?

DoilLee2222: Not separate logic — just comment syntax. The rule is the same in both languages (mark the offending line in a negative fixture with a `VIOLATION` comment); only the token differs: `// VIOLATION: <rule-key>` for TS/JS, `# VIOLATION: <rule-key>` for Python.

Python is in scope because `fp-next-fix` writes Python fixtures (steps 5–6 target `sample-python-project-*`), so a batch PR can contain one and this review must accept its `#` marker. It also mirrors `fp-next-fix.md` step 6, which already says "the language-appropriate comment for Python."

---

mushgev: how is it for C# ?

It's a surpirse for me that it is different. Can you ask Claude to fix this in a separate local session, to make it // VIOLATION

DoilLee2222: C# is **not** different — it already uses `// VIOLATION: <rule-key>`, same as TS/JS. C# comments are `//`, so it's covered by the default case with no carve-out.

Python is the only exception, and it's unavoidable: Python has no `//` comment syntax at all — its only line-comment token is `#`. So a Python fixture physically can't use `// VIOLATION`; it must be `# VIOLATION: <rule-key>`. That's a language constraint, not a choice we can normalize away.

So of the three fixture languages: TS/JS → `//`, C# → `//`, Python → `#`. Nothing to fix for C#; the Python note is the only special case, and it has to stay.

---

mushgev: Why it has to stay? Why we can't make it same for Python as well?

DoilLee2222: Because `//` isn't a comment in Python — it's the floor-division operator. A `.py` file containing `// VIOLATION: <rule-key>` is a **syntax error** and won't parse, and the analyzer parses these fixtures. So we physically can't write the marker with `//` in Python.

It's also baked into the test harness, which scans for each language's own comment token:
- `tests/analyzer/js-negative.test.ts:56` → `/\/\/\s*VIOLATION:\s*(.+)/` (matches `//`)
- `tests/analyzer/python-negative.test.ts:51` → `/#\s*VIOLATION:\s*(.+)/` (matches `#`)

So making Python use `//` would require both (a) valid Python that `//` can't provide, and (b) changing the Python harness regex — with no benefit, since the marker is just "a comment on the offending line" and each language already uses its native comment syntax. That's why the `#` note stays.

---

mushgev: Makes sense!

Now you can test this before merging.

PR #737, opened 2026-07-09
fix(fp): stop campaign-stuck tracker from double-notifying to Telegram

### LINKED ISSUE #735: [cs-fp-campaign-stuck] abpframework/abp
The `abpframework/abp` C# campaign cannot self-progress through the automated fp-next-fix routine. This is the human-signal issue; details below.

## Where the campaign stands

Queue-empty re-measurement this session (fresh `pnpm build:dist`, `node dist/cli.mjs analyze --no-llm --no-stash --no-skills` at `8665ce0a23622f658b531b0627c7c025935fdb3b`):

- **total_violations:** 76,196 across 556 rules
- **Sampled tp_rate ≈ 0.92** — classified ≤10 samples/rule across the top 80 rules (~87% of all violations, 748 samples): **674 TP / 55 FP / 19 borderline**. Counting every borderline as FP still gives 0.901. So it nominally clears the 0.90 gate — but see the two problems below.

## Problem 1 — a prior campaign-close was already human-rejected

PR **#690** (`chore(fp): close abpframework/abp campaign, bump to v0.6.11`, claimed tp_rate 0.97 over a **C#-only** top-50 sample) was **closed unmerged on 2026-07-03**. The campaign then continued (fix PRs #714, #718, #719, #723, #724, #729 merged through 2026-07-08). Re-opening a campaign-close now — especially at a *lower*, more-complete 0.92 that includes the FP-heavy JS rules #690 excluded — would just duplicate rejected work, so this session did **not** open one.

## Problem 2 — the residual FP-bug rules can't be fixed within the routine's edit scope

The more-complete measurement surfaces genuine detection bugs (verified against source), mostly in the JS/TS rules that fire on abp's `wwwroot` frontend and vendored libraries:

| rule_key | hits | sampled FP rate | FP shape |
|---|---:|---|---|
| `code-quality/deterministic/class-prototype-assignment` | 138 | ~100% | fires on ES5 constructor-function prototypes (no ES6 `class` present) — see #734 |
| `bugs/deterministic/no-undef` | 187 | ~90% | flags JS built-ins (`arguments`, `self`, `getComputedStyle`) and script-loaded globals (`$`, `jQuery`, `abp`, `bootstrap`, `luxon`) |
| `code-quality/deterministic/block-scoped-var` | 255 | ~60% | flags `var` used only within its own block, wrongly claiming cross-block use |
| `code-quality/deterministic/unsafe-any-usage` | 627 | ~60% | weak TS type resolution: `as`-casts, `inject<T>()`, imported typed consts, lib types reported as `any` |
| `code-quality/deterministic/max-nesting-depth` | 655 | ~40% | counts `else if` chains as cumulative nesting depth |
| `code-quality/deterministic/dead-store` | 133 | ~40% | flags `var x, y;` declaration lists |
| `code-quality/deterministic/for-in-without-filter` | 206 | ~40% | doesn't recognize a hoisted `hasOwnProperty` guard as a filter |
| `code-quality/deterministic/no-extraneous-class` | 199 | ~30% | flags Angular `@NgModule` classes (`static forRoot`) that can't be static |
| `code-quality/deterministic/commented-out-code` | 197 | ~40% | flags prose / JSDoc / TODO notes as commented-out code |
| `code-quality/deterministic/unused-expression` | 115 | ~30% | flags IIFEs / UMD wrappers |
| `code-quality/deterministic/missing-return-type` | 906 | ~2% | fires on TS `set` accessors, which can't carry a return type |

**The blocker:** each of these is a legitimate visitor fix, but the corresponding unit tests in `tests/analyzer/*.test.ts` encode the current (buggy) behavior, and those files are **outside the fp-fix edit scope** (`packages/analyzer/src/rules/`, `.../patterns/`, `tests/fixtures/sample-*/` only). Concretely, fixing `class-prototype-assignment` breaks `tests/analyzer/code-quality-rules.test.ts:2584`, which asserts an ES5-constructor prototype assignment *should* fire (see #734). The fp-next-fix-review routine also rejects any PR touching out-of-scope files. So a human needs to land each visitor fix **together with** its unit-test update.

Note: `code-quality/deterministic/identifier-matches-keyword` (193 hits) is **not** a bug — it is CA1716 (cross-CLR-language keyword collisions), working as designed; #690's body mislabeled it as FP.

## Blocked issues needing human resolution

- [ ] #734 — `class-prototype-assignment` (fix ready; needs the out-of-scope unit-test update to land)

The other rules above are not yet filed as individual issues because they share the same structural blocker; filing them would only produce more `cs-fp-blocked` duplicates. Happy to file them per-rule if that's preferred.

## Environment notes (for whoever picks this up)

1. **`pnpm test` needs a full `pnpm build` first** in this environment — the routine setup only runs `pnpm build:dist`, so `@truecourse/ee-db` (and the other `ee/packages/*`) have no `dist/` entry and 47 `ee-*`/store/`github-app` test files fail to load with `Failed to resolve entry for package "@truecourse/ee-db"`. Unrelated to any rule change.
2. **C# analyze needs `DOTNET_ROLL_FORWARD=LatestMajor`** here. `setup-csharp-env.sh` installs the .NET 10 SDK **and** a .NET 8 runtime; the net8-targeted Roslyn host then runs on net8 (lowest satisfying `rollForward: Major`) and MSBuildLocator can't load the net10 SDK's MSBuild from a net8 process → `No instances of MSBuild could be detected`, which fails the analyze hard. Forcing the host onto net10 via `DOTNET_ROLL_FORWARD=LatestMajor` (matching the documented "runs on any ≥ 8 runtime with roll-forward" design) fixes it. Worth baking into the cs- environment so every session doesn't rediscover it.

cc @mushgev

DoilLee2222:
Surfaced by #735. This PR fixes the **double-Telegram-notification bug** that #735 exposed. It does **not** resolve #735's underlying campaign-stuck condition (the abp blocked-rule fixes) — that stays open.

## Problem

Issue #735 (`[cs-fp-campaign-stuck] abpframework/abp`) sent **two** Telegram messages on open instead of one:

| Message | Job in `notify-routine-activity.yml` | Fired because |
|---|---|---|
| ⚠️ Campaign needs attention: #735 | `notify-campaign-alert` | **Title** contains `fp-campaign-stuck]` ✅ intended |
| 📋 1 new cs-fp-fix issue filed • #735 | `notify-fix-burst` | Issue carried the **`cs-fp-fix` label** ❌ spurious |

## Root cause

The campaign-stuck tracker is a campaign-level human-signal issue, not an FP-fix queue item — its alert should come only from `notify-campaign-alert` (keyed on the title tag). The queue-empty path in `fp-next-fix.md` never specified the tracker's labels, so the routine defaulted to the campaign's fix-queue label `cs-fp-fix`, which swept the tracker into the `notify-fix-burst` rollup.

The same mislabel had a second, latent effect: `cs-fp-fix` makes the tracker match step 1's `label:cs-fp-fix` search, so the next `fp-next-fix` run would have picked it up as a fixable FP issue, commented "malformed YAML", and blocked it.

`fp-stale-pr-notify` already avoids this by giving its tracker a dedicated `fp-stale-pr` label — this change brings the campaign-stuck tracker in line, and hardens both consumers (notifier + fix loop) against a future mislabel.

## Changes

**Root cause — don't mislabel the tracker**
- `docs/fp-automation/prompts/fp-next-fix.md` / `docs/drift-fp-automation/prompts/drift-fp-next-fix.md`: the campaign-stuck tracker is labelled `fp-target:<owner>-<repo>` **only**, never the fix-queue label.

**Defence-in-depth — notifier ignores mislabeled trackers**
- `.github/workflows/notify-routine-activity.yml`: the fix-burst rollup filters trackers out of its count by matching the **exact tracker title tags** the sibling notifier jobs key on (`fp-campaign-close]` / `fp-campaign-stuck]` / `drift-campaign-broken]` / `fp-stale-pr]`), anchored on the closing `]`. (An earlier loose `stale` substring would have wrongly suppressed real rollups for rule keys like `database/llm/stale-read-after-write`.)

**Defence-in-depth — fix loop ignores mislabeled trackers**
- `fp-next-fix.md` / `drift-fp-next-fix.md` step 1: exclude any issue whose title contains a tracker tag from the pickup search, so a stray fix label can't make a session grab a tracker and block it.

**Docs**
- `docs/fp-automation/README.md`: document the label constraint.

## Out-of-band cleanup (not in this diff)

Removed the erroneous `cs-fp-fix` label from the live issue #735, so it now carries `fp-target:abpframework-abp` only and won't be re-picked by the fix loop.

cc @mushgev

### DIFF
--- a/.github/workflows/notify-routine-activity.yml
+++ b/.github/workflows/notify-routine-activity.yml
@@ -105,13 +105,21 @@ jobs:
           fi
 
           since=$(date -u -d '10 minutes ago' '+%Y-%m-%dT%H:%M:%SZ')
+          # Defence-in-depth: a campaign-level tracker ([<SCOPE>fp-campaign-stuck],
+          # [<SCOPE>fp-stale-pr], …) is handled by notify-campaign-alert /
+          # notify-stale-pr and must never appear in this fix-burst rollup. It
+          # only could if a routine mislabels it with a fix label (the
+          # fp-next-fix / drift-fp-next-fix prompts forbid this). Drop any such
+          # title here so a stray label can't double-notify: a genuine fix issue
+          # is titled [<SCOPE>fp-<rule-key>] and never contains these tokens.
           issues_json=$(gh issue list \
             --repo "$REPO" \
             --label "$LABEL" \
             --state open \
             --search "created:>${since}" \
             --json number,title \
-            --limit 200)
+            --limit 200 \
+            | jq '[.[] | select(.title | test("campaign-|stale-pr") | not)]')
 
           count=$(echo "$issues_json" | jq 'length')
           if [ "$count" -eq 0 ]; then
--- a/.github/workflows/notify-routine-activity.yml
+++ b/.github/workflows/notify-routine-activity.yml
@@ -105,13 +105,16 @@ jobs:
           fi
 
           since=$(date -u -d '10 minutes ago' '+%Y-%m-%dT%H:%M:%SZ')
-          # Defence-in-depth: a campaign-level tracker ([<SCOPE>fp-campaign-stuck],
-          # [<SCOPE>fp-stale-pr], …) is handled by notify-campaign-alert /
-          # notify-stale-pr and must never appear in this fix-burst rollup. It
-          # only could if a routine mislabels it with a fix label (the
-          # fp-next-fix / drift-fp-next-fix prompts forbid this). Drop any such
-          # title here so a stray label can't double-notify: a genuine fix issue
-          # is titled [<SCOPE>fp-<rule-key>] and never contains these tokens.
+          # Defence-in-depth: a campaign-level tracker is handled by
+          # notify-campaign-alert / notify-stale-pr and must never appear in
+          # this fix-burst rollup. It only could if a routine mislabels it with
+          # a fix label (the fp-next-fix / drift-fp-next-fix prompts forbid
+          # this). Drop any tracker title here so a stray label can't
+          # double-notify. Match the exact title tags the sibling notifier jobs
+          # key on (anchored on the closing `]`) rather than loose substrings —
+          # a genuine fix issue is titled [<SCOPE>fp-<rule-key>] and some rule
+          # keys legitimately contain "stale" (e.g. .../stale-read-after-write),
+          # so a bare "stale" match would wrongly suppress real rollups.
           issues_json=$(gh issue list \
             --repo "$REPO" \
             --label "$LABEL" \
@@ -119,7 +122,7 @@ jobs:
             --search "created:>${since}" \
             --json number,title \
             --limit 200 \
-            | jq '[.[] | select(.title | test("campaign-|stale-pr") | not)]')
+            | jq '[.[] | select(.title | test("fp-campaign-close]|fp-campaign-stuck]|drift-campaign-broken]|fp-stale-pr]") | not)]')
 
           count=$(echo "$issues_json" | jq 'length')
           if [ "$count" -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


DoilLee2222:
Closing: the fp-automation program is retired in favour of the agentic-loop approach.

This PR only fixed the campaign tracker's double-Telegram-notification (surfaced by #735, itself now closed as not planned). With the fp routines retired there is no tracker left to notify, so the fix has nothing to apply to. Not a comment on its correctness — it just has no subject any more.

PR #738, opened 2026-07-09
Optimize fp-next-fix routines: work the human-review queue instead of parking it

DoilLee2222:
## What & why

Today `fp-next-fix` treats a hard-to-fix false positive as a dead end: it labels the issue `fp-blocked` and, once only blocked issues remain, files a "needs human input" tracking issue and stops. Since we're now using Claude to work those harder fixes too, this PR turns that dead-end into a **worked queue**.

## Changes

**Label rename:** `fp-blocked` → `fp-human-review-needed` (scoped: default `fp-human-review-needed`, C# account `cs-fp-human-review-needed`).

**`fp-next-fix` — two phases, one PR per run** (the "at most one PR per session" constraint is preserved). The phase is fixed at session start by which queue has pickable work:

- **Normal phase** (default): fix up to 5 issues that **don't** carry `fp-human-review-needed`, using only **in-bounds** changes (the rule's visitor/pattern + fixtures). Opens a normal PR (label `fp-fix`) that `fp-next-fix-review` may auto-merge. Fewer than 5 pickable → fix that many. When a fix would need to cross the in-bounds boundary (cross-module refactor, or the rule is broken beyond the FP), the issue is **deferred** with `fp-human-review-needed` instead of being dead-blocked.
- **Human-review phase**: entered only once the normal queue is empty but `fp-human-review-needed` issues remain. The session is allowed to **go over the boundary** (targeted cross-module refactors) to fix up to 5 of them, and opens a PR carrying **both** `fp-fix` and `fp-human-review-needed` (plus a `## Human review required` body section listing what went over the boundary).

**Terminal handling retained:** issues that can't be fixed even with out-of-bounds access (malformed issue data, an infeasible refactor) are marked `fp-skipped` and surfaced to a human via the retained `[fp-campaign-stuck]` tracking issue. Bail-out routing is spelled out explicitly (defer vs terminal vs close).

**`fp-next-fix-review` — reviews every PR, merges only the safe ones:**

- Reviews **every** batch PR regardless of label.
- **Normal PR:** unchanged — compliant → merge (fires the next session); non-compliant → open a `[fp-review-failed]` issue, don't merge.
- **`fp-human-review-needed` PR:** **never merges.** It posts its review as a PR comment to help the human and stops — no review-failed issue (the human is already in the loop). A human reviews and merges.
- The in-bounds/scoped checks (checks 1 & 5) are **phase-aware**: cross-boundary changes are expected on a human-review PR and aren't flagged, while touching `campaigns.yaml` or the four version-bump locations stays a hard violation on either PR kind.

**Docs:** `docs/fp-automation/README.md` updated to match (label list, issue-status enum, per-issue-loop summary, queue-empty/stuck path, two-phase description).

## Not touched (by design)

- `fp-stale-pr-notify` — it keys on the presence of the `fp-reviewed` label, which the review routine now also adds to human-review PRs after reviewing, so those aren't falsely flagged as stuck. Functionally correct as-is; left out of this PR's two-prompt scope.
- The separate `drift-fp-automation` chain — a parallel system, not in scope.

cc @mushgev

---
_Generated by [Claude Code](https://claude.ai/code/session_01EtZ8sQ7ZL6yJeqBMwicJxR)_

### DIFF
--- a/docs/fp-automation/prompts/fp-next-fix-review.md
+++ b/docs/fp-automation/prompts/fp-next-fix-review.md
@@ -7,27 +7,28 @@ produced by the `fp-next-fix` routine — the test fixtures and the
 visitor/pattern edits — decide whether **that code** faithfully follows
 every code-level instruction in
 [`docs/fp-automation/prompts/fp-next-fix.md`](./fp-next-fix.md), and then
-take exactly one terminal action:
+take exactly one terminal action.
 
-- **Compliant** → **merge the PR**. The merge fires the next
-  `fp-next-fix` session (Trigger B), so the inner loop keeps running
-  with no human in the loop.
-- **Non-compliant** → **do not merge**. Open one issue on
-  `truecourse-ai/truecourse` describing every violation, and comment
-  the same summary on the PR. (The issue is the human signal; leave the
-  PR open for a human to fix or close.)
+**You review every fp-fix batch PR regardless of its labels.** But the
+`<SCOPE>fp-human-review-needed` label on the PR changes what you're
+allowed to do with it:
 
-**Scope: code only.** This routine reviews *the committed code* — the
-fixtures under `tests/fixtures/sample-*/…` and the visitor/pattern edits
-under `packages/analyzer/src/…`. It does **not** review the PR's
-structure: the title format, the PR-body markdown sections
-(`Closes #…` lines, the `## Fixes` table, per-rule sections, the
-`## FP-count delta` write-up), the PR/issue labels, the linked-issue
-bookkeeping, and commit/PR hygiene trailers are all **out of scope** and
-must **never** produce a violation here. A PR whose code is correct is
-merged even if its body or title is imperfect; a PR whose code is wrong
-is failed even if its body is immaculate. (PR structure is fp-next-fix's
-own responsibility and is checked elsewhere.)
+- **Normal PR** (no `<SCOPE>fp-human-review-needed` label) — the fix is
+  in-bounds and auto-mergeable:
+  - **Compliant** → **merge the PR**. The merge fires the next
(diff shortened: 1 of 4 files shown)


docs/fp-automation/prompts/fp-next-fix-review.md
```diff
@@ -92,7 +100,7 @@ point at concrete evidence in the diff or CI. Do **not** invent problems;
 a requirement you cannot evaluate is noted as `unverifiable: <reason>`
 rather than a violation. **Do not** append violations for anything about
 the PR's title, body, labels, linked-issue state, or commit trailers —
-those are out of scope (see "Scope: code only" above).
+those are out of scope (see "Scope: code only" below).
```
spiderhands: There is no "Scope: code only" block below. It was removed _above_, but never added _below_

docs/fp-automation/prompts/fp-next-fix-review.md
spiderhands: As our new workflow will always fail the check for human reviewable fixes, there always would be violations. This should be aware of the new flow

docs/fp-automation/prompts/fp-next-fix-review.md
DoilLee2222: fixed in c1c2aef. Removed the dead reference and pointed it at the Code only bullet under Hard constraints instead.

docs/fp-automation/prompts/fp-next-fix-review.md
DoilLee2222: Fixed in c1c2aef. Check 6 now excludes the block-human-review gate — it's a merge gate, not a test, so it's expected to be red on human-review PRs and the routine no longer counts it as a failed check or violation. Only real CI checks (tests/build) are evaluated.

spiderhands (approved):

PR #739, opened 2026-07-10
fix(fp): resolve 1 FP from abpframework/abp

### LINKED ISSUE #734: [cs-fp-fix] code-quality/deterministic/class-prototype-assignment in abpframework/abp
```yaml
target_repo: abpframework/abp
target_ref: 8665ce0a23622f658b531b0627c7c025935fdb3b
rule_key: code-quality/deterministic/class-prototype-assignment
fp_count: 10
samples:
  - url: ``https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/toast/abp-toast.js#L43``
    why_fp: The receiver `AbpToastService` is an ES5 constructor function (`function AbpToastService(){}`), not an ES6 class, so `AbpToastService.prototype.method = function(){}` is the correct/idiomatic ES5 pattern — there is no "class context" to be inconsistent with.
  - url: https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/index.js#L406
    why_fp: Same shape — prototype method assignment on a plain constructor function, no `class` declaration anywhere for the receiver.
  - url: https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/npm/packs/jquery/src/abp.jquery.js#L300
    why_fp: Classic ES5 prototype-based library code; the rule's "use class method definitions instead" advice does not apply because the type is not a class.
status: open
pr: null
```

## Borderline cases

None. The rule (`class-prototype-assignment`) is intended to catch the *inconsistent* pattern of declaring an ES6 `class` and then assigning methods onto its `.prototype` externally. The visitor currently fires on **any** `X.prototype.method = function(){}` regardless of whether `X` is an ES6 class, so it misfires on ordinary ES5 constructor-function code (138 hits across 20 files in this target, all ES5). Fix: only flag when the receiver identifier is actually declared as a `class` in the same file.

---
_Filed by the fp-next-fix queue-empty re-measurement (SCOPE=cs-) of `abpframework/abp` @ `8665ce0a23622f658b531b0627c7c025935fdb3b`. Engine: tree-sitter JS/TS visitor._

DoilLee2222:
## Human review required

This PR was produced in the **human-review phase** of `fp-next-fix` (SCOPE=`cs-`): the only pickable `cs-fp-fix` issue, #734, carried `cs-fp-human-review-needed`, so this batch is reviewed by `fp-next-fix-review` but **merged by a human**.

What to focus on:
- The rule fix itself is **in-bounds** (only edits the rule visitor under `packages/analyzer/src/rules/…` + fixtures).
- One change goes **over the normal in-bounds boundary**: `tests/analyzer/code-quality-rules.test.ts`. The existing `detects prototype method assignment` unit test asserted the **pre-fix (buggy)** behavior — that an ES5 constructor's `.prototype` assignment fires — which is exactly the false positive #734 reports. It's updated to the corrected semantics (fire on an ES6 `class` receiver) plus a regression guard for the ES5 case. This test edit is why the PR is human-merged.

Closes #734

## Fixes

| rule_key | issue | positive-fixture | negative-fixture |
|---|---|---|---|
| `code-quality/deterministic/class-prototype-assignment` | #734 | `sample-js-project-positive/class-prototype-assignment-es5-constructor.js` | `sample-js-project-negative/class-prototype-assignment-es6-class.js` |

## code-quality/deterministic/class-prototype-assignment (#734)

OSS sources (abp @ `8665ce0a23622f658b531b0627c7c025935fdb3b`):
- ``https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/framework/src/Volo.Abp.AspNetCore.Mvc.UI.Theme.Shared/wwwroot/libs/abp/aspnetcore-mvc-ui-theme-shared/toast/abp-toast.js#L43``
- https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/modules/docs/src/Volo.Docs.Web/Pages/Documents/Project/index.js#L406
- https://github.com/abpframework/abp/blob/8665ce0a23622f658b531b0627c7c025935fdb3b/npm/packs/jquery/src/abp.jquery.js#L300

Summary: the rule flagged **any** `X.prototype.method = function(){}` regardless of whether `X` is an ES6 `class`, so it misfired on ordinary ES5 constructor-function code (abp's idiom). Added a `collectClassNames` helper that walks the file AST for `class Foo {}` declarations and `const Foo = class {}` expressions; the rule now fires only when the `.prototype` receiver is a **simple identifier declared as a class in the same file**. Non-class (builtin or ES5 constructor) receivers are skipped. Grafting a method onto a real class's prototype — the actual "inconsistent with class syntax" pattern — still fires (see negative fixture).

## FP-count delta (vs `8665ce0a23622f658b531b0627c7c025935fdb3b` on `abpframework/abp`)

| Rule | Before | After | Delta |
|---|---:|---:|---:|
| `code-quality/deterministic/class-prototype-assignment` | 138 | 0 | -138 |
| **Total (these 1 rules)** | **138** | **0** | **-138** |
| **All-rules total on target** | — | — | `unavailable: see note` |

**Measurement note.** The standard `node dist/cli.mjs analyze --no-llm` could not complete on this session: the project-aware **C# Roslyn tier fail-hard'd** loading `framework/Volo.Abp.slnx` (`No instances of MSBuild could be detected` — a net8 host / SDK-10 MSBuild mismatch), even after `dotnet restore` and installing the .NET 10 SDK. That aborts the whole run before results are saved, so the all-rules total is unavailable. This is an environment/toolchain limitation and is **unrelated to this rule**, which is a pure tree-sitter JS rule. The per-rule delta above was measured by running only this rule over abp's analyzer-discovered `.js` files (299 files — the same set the real analyze scans): **138 → 0**, exactly matching the count in #734. The full JS/TS analyzer suite is green (`js-positive`, `js-negative`, `code-quality-rules`).

cc @mushgev

### DIFF
--- a/packages/analyzer/src/rules/code-quality/visitors/javascript/class-prototype-assignment.ts
+++ b/packages/analyzer/src/rules/code-quality/visitors/javascript/class-prototype-assignment.ts
@@ -1,6 +1,35 @@
 import type { CodeRuleVisitor } from '../../../types.js'
 import { makeViolation } from '../../../types.js'
 
+// Collect the names bound to an ES6 `class` in this file — both
+// `class Foo {}` declarations and `const Foo = class {}` expressions. The rule
+// only makes sense when the `.prototype` receiver is actually one of these:
+// grafting a method onto a real class's prototype is inconsistent with class
+// syntax, whereas the same shape on an ES5 constructor function is idiomatic.
+function collectClassNames(root: import('web-tree-sitter').Node): Set<string> {
+  const names = new Set<string>()
+  const stack: (import('web-tree-sitter').Node)[] = [root]
+  while (stack.length > 0) {
+    const n = stack.pop()!
+    if (n.type === 'class_declaration') {
+      const name = n.childForFieldName('name')
+      if (name?.text) names.add(name.text)
+    } else if (n.type === 'variable_declarator') {
+      const name = n.childForFieldName('name')
+      const value = n.childForFieldName('value')
+      if (name?.type === 'identifier' && value &&
+          (value.type === 'class' || value.type === 'class_expression')) {
+        names.add(name.text)
+      }
+    }
+    for (let i = 0; i < n.namedChildCount; i++) {
+      const c = n.namedChild(i)
+      if (c) stack.push(c)
+    }
+  }
+  return names
+}
+
 export const classPrototypeAssignmentVisitor: CodeRuleVisitor = {
   ruleKey: 'code-quality/deterministic/class-prototype-assignment',
   languages: ['typescript', 'tsx', 'javascript'],
@@ -19,10 +48,25 @@ export const classPrototypeAssignmentVisitor: CodeRuleVisitor = {
 
     const right = node.childForFieldName('right')
     if (right?.type === 'function_expression' || right?.type === 'arrow_function') {
-      const builtinObj = obj.childForFieldName('object')
+      const receiver = obj.childForFieldName('object')
+
+      // A simple identifier receiver is required to reason about whether it is
+      // a class. `X.prototype.m = fn` where `X` is anything else (a member
+      // expression, a call, etc.) is not the ES6-class inconsistency this rule
+      // targets.
+      if (receiver?.type !== 'identifier') return null
+
       const BUILTINS = new Set(['Array', 'Object', 'String', 'Number', 'Boolean', 'Function',
         'RegExp', 'Date', 'Error', 'Map', 'Set', 'Promise', 'Symbol'])
-      if (builtinObj?.type === 'identifier' && BUILTINS.has(builtinObj.text)) return null
+      if (BUILTINS.has(receiver.text)) return null
+
+      // Only fire when the receiver is actually declared as an ES6 class in the
+      // same file. Assigning `.prototype` methods on a plain ES5 constructor
+      // function is idiomatic ES5, not an inconsistency with class syntax.
+      let root = node
+      while (root.parent) root = root.parent
+      const classNames = collectClassNames(root)
+      if (!classNames.has(receiver.text)) return null
 
       return makeViolation(
         this.ruleKey, node, filePath, 'low',
--- a/tests/analyzer/code-quality-rules.test.ts
+++ b/tests/analyzer/code-quality-rules.test.ts
@@ -2581,13 +2581,24 @@ describe('code-quality/deterministic/require-yield', () => {
 });
 
 describe('code-quality/deterministic/class-prototype-assignment', () => {
-  it('detects prototype method assignment', () => {
+  it('detects prototype method assignment on an ES6 class', () => {
+    const violations = check(`
+      class Animal { constructor(name) { this.name = name; } }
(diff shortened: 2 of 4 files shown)


DoilLee2222:
**fp-next-fix-review** (SCOPE=`cs-`): code review **passed**. Held for human merge per `cs-fp-human-review-needed` — this routine does not merge human-review PRs.

Reviewed the committed code against the code-level instructions in `fp-next-fix.md`. `violations = []`.

Checklist (derived fixed set from the diff's `// VIOLATION:` markers = `code-quality/deterministic/class-prototype-assignment`, 1 rule):
- ✅ **Files in-bounds** — the rule visitor + both `sample-*` fixtures are in-bounds; `tests/analyzer/code-quality-rules.test.ts` is over the normal boundary but expected on a human-review PR (it is not `campaigns.yaml` nor a version-bump location).
- ✅ **Fix count** — 1 rule (≥1, ≤5).
- ✅ **Fixtures** — positive (`…-es5-constructor.js`) has no `// VIOLATION:`; negative (`…-es6-class.js`) marks the offending line with `// VIOLATION: code-quality/deterministic/class-prototype-assignment`; both are generic paraphrases with no upstream owner/repo/filename identity.
- ✅ **Complete** — positive + negative fixture + a visitor change for the rule.
- ✅ **Scoped** — every non-fixture change traces to the fixed rule; no unrelated refactor.
- ✅ **CI** — the `test` check is green.

Cross-boundary change (the reason a human merges):
- `tests/analyzer/code-quality-rules.test.ts` — the existing `detects prototype method assignment` unit test asserted the **pre-fix (buggy)** behavior (an ES5 constructor's `.prototype` assignment firing). It's updated to fire on an ES6 `class` receiver, plus a new regression guard for the ES5 case. Targeted and sound — it corrects the test that codified the FP, nothing more.

Review notes (non-blocking):
- The fix logic reads correctly: `collectClassNames` walks the file AST for `class Foo {}` and `const Foo = class {}`, and the rule now only fires when the `.prototype` receiver is a simple identifier that is a class in the same file (builtins and ES5 constructors are skipped). The negative fixture confirms the ES6-class case still fires; the positive confirms ES5 no longer does.
- **Minor / optional:** the fixture identifiers (`Toast` / `ToastService`) echo the *domain theme* of one upstream sample file (a toast component). It does not leak the project/owner/repo/filenames, so it is not a check-3 violation, but a fully domain-neutral name (e.g. `Widget` / `WidgetService`) would be marginally safer for the public corpus.

No `[cs-fp-review-failed]` issue opened — per the human-review-PR path, the human is already the reviewer/merger. Safe to merge.

cc @mushgev

---

PR #755, opened 2026-07-14
fix(guard): make the package-under-test resolvable by name in scenario sandboxes

### LINKED ISSUE #754: guard birth-validation: scenarios importing the package-under-test by name fail with MODULE_NOT_FOUND (false negatives)
## Summary

During `guard generate`, birth-validation runs each generated scenario in an isolated temp work dir (`/private/.../T/tc-guard-XXXX/work/`). That sandbox has **no `node_modules/<package-under-test>`**, so any scenario whose setup imports the package by its public name (`import 'tsx'`, `require('tsx/cjs/api')`, `from 'tsx/esm/api'`) fails at **module resolution — before any behavior of the package is exercised**. These scenarios are then discarded as birth findings.

The result is a class of **false negatives**: valid, useful scenarios for a package's programmatic API get thrown away for an environment reason, and guard becomes blind to that entire API surface.

## Where it was observed

- Package/CLI: `truecourse@0.7.1`, Node `v24.14.0`
- Target repo under test: [`privatenumber/tsx`](https://github.com/privatenumber/tsx) (a deterministic JS/TS library)
- Pipeline: `spec scan` → `guard generate` → `guard run`

`guard generate` result for tsx: `sectionsTotal: 84`, `written: 45`, `birthPassed: 51`, **`birthFindings: 20`**, `coverageGaps: 57`.

**18 of the 20 birth findings are this bug** (the other 2 are unrelated resolution-order scenarios — see Notes).

## Evidence

Three pieces line up to show the failure is resolution, not behavior:

**1. How the scenario is run** (`invocation.json` → `steps[0].argv`), from an ephemeral temp dir:
```
node <repo>/tsx/dist/cli.mjs entry.mjs
```

**2. What the scenario setup does** (generated YAML):
```yaml
setup:
  files:
    entry.mjs: |
      import 'tsx'
      const mod = await import('./file.ts')
      console.log('value:', mod.value)
```

**3. What happens** (`stderr.txt`):
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'tsx'
  imported from /private/.../T/tc-guard-QiTt0C/work/entry.mjs
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    ...
    at resolveTsPathsSync (file://<repo>/tsx/dist/register-*.mjs)
  code: 'ERR_MODULE_NOT_FOUND'
```
```
Error: Cannot find module 'tsx/cjs/api'
Require stack:
- /private/.../T/tc-guard-4CnPGH/work/register.cjs
  code: 'MODULE_NOT_FOUND'
```

Note the tsx loader is actually active in the stack (`resolveTsPathsSync` in `tsx/dist/register-*.mjs`) — the CLI injected it — but resolving the **bare specifier** `tsx` from the sandbox still fails because the package isn't linked into that work dir.

**Aggregate MODULE_NOT_FOUND specifiers across the 20 findings' evidence:**

| Missing specifier | occurrences |
|---|---|
| `Cannot find module 'tsx/cjs/api'` | 14 |
| `Cannot find package 'tsx'` | 13 |
| fallback path attempts (`.../work/dist/cjs/api/index.cjs`, `'tsx/cjs'`, `dist/esm/api/index.mjs`) | ~8 |
| `./target.js`, `./mod.js` (the 2 unrelated scenarios) | 2 |

## Root cause

Birth-validation does not make the **package-under-test resolvable** inside the scenario work dir. Scenarios that exercise the programmatic API (register/require/tsImport) must import the package by its public name, so Node throws `MODULE_NOT_FOUND` at resolution time and the scenario is recorded as a (spurious) birth finding.

## Impact

For tsx this discarded **every scenario for the entire `docs/dev-api/*` surface** — `register-cjs`, `register-esm`, `tsx-require`, `entry-point` (`tsx.register()`, scoped `register({ namespace })`, `tsx/cjs/api`, `tsx.require()`, `onImport`, `unregister()`, `import 'tsx'`). Guard therefore has **no coverage of the programmatic API** — which is exactly the surface where such libraries tend to have real bugs. Any package that documents a programmatic/import-by-name API will hit this.

## Suggested fix

Make the package-under-test resolvable in the scenario sandbox before birth-validation, e.g. one of:
- symlink `node_modules/<pkg>` → the repo root (the repo *is* the package), or
- `npm link` / install the local package into the work dir, or
- rewrite bare specifiers in generated scenarios to the built entry path.

Once `tsx` / `tsx/cjs/api` / `tsx/esm/api` resolve in the work dir, these ~18 scenarios should execute and become real behavioral tests of the API.

## Notes

- The other **2 of 20** findings are unrelated CJS resolution-order scenarios (`./x.js` must not resolve `x.js.ts`; TS candidate should be preferred over an existing `.js`). Those may be mis-generated expectations or genuine resolution nuances — tracked separately, not part of this bug.

## Environment

- `truecourse@0.7.1`
- Node `v24.14.0`, macOS (darwin 24.6.0)
- Target: `privatenumber/tsx` @ commit `5828ffe`

mushgev:
Fixes #754

## Problem

During `guard generate`, birth-validation runs each scenario in a bare temp work dir. Any scenario whose `setup.files` import the package-under-test by its published name (`import 'tsx'`, `require('tsx/cjs/api')`) — exactly how the docs tell users to consume it — died at Node module resolution (`MODULE_NOT_FOUND`) before any behavior ran, and was discarded as a spurious birth finding. For tsx this threw away the entire `docs/dev-api/*` surface (18 of 20 birth findings), leaving guard blind to the programmatic API.

## Fix

`npm link` semantics, applied per sandbox:

- **`package-links.ts`** (new): `discoverPackageLinks(repoRoot)` finds the repo-root package by name plus, for monorepos, every workspace package — npm/yarn `workspaces` (array and `{ packages }` forms) and `pnpm-workspace.yaml`, with `*`/`**` glob and `!` exclusion support. Non-Node repos yield an empty list (no-op).
- **`sandbox.ts`**: each discovered package is symlinked into the sandbox at `node_modules/<name>` (junction type — privilege-free on Windows). Node resolves through the link's realpath, so the package's own dependencies resolve from the repo's real `node_modules` and subpaths (`tsx/cjs/api`) resolve through its actual `exports` map.
- **`run.ts` / `run-scenario.ts`**: discovery runs once per `runGuard` and reaches every sandbox. Birth-validation calls `runGuard` too, so birth and real `guard run` are fixed by the same path.

Safety: links are created **after** `setup.files` seeding and never over an existing path — a scenario that seeds its own `node_modules/<name>` stub keeps it, and seeded files can never be written *through* a link into the real repo. A link name that would escape `node_modules` throws `SandboxError`.

## Tests

- 11 discovery unit tests (`package-links.test.ts`): root/scoped names, both workspace forms, pnpm yaml + `!` exclusions, `**` traversal guards, dedup, malformed inputs, escape rejection.
- 4 sandbox tests: link creation, scoped parent dirs, seeded-stub-wins, escape rejection.
- 2 end-to-end regressions in `run.test.ts`: the exact #754 failure mode (bare-name ESM import + CJS subpath require + a dependency that only resolves through the link's realpath), and a pnpm-workspace monorepo importing `@acme/lib`.

Full suite: 288 files / 7360 tests green.

The other 2 of 20 tsx findings (CJS resolution-order scenarios) are unrelated and tracked separately per the issue.

https://claude.ai/code/session_016U8xzfBwVuuHafRPpug6P5

### DIFF
--- a/packages/guard-runner/src/index.ts
+++ b/packages/guard-runner/src/index.ts
@@ -25,6 +25,9 @@ export type { RunScenarioContext } from './run-scenario.js'
 export { createSandbox, SandboxError, listSandboxFiles, DETERMINISM_PINS } from './sandbox.js'
 export type { Sandbox, SandboxOptions } from './sandbox.js'
 
+export { discoverPackageLinks } from './package-links.js'
+export type { PackageLink } from './package-links.js'
+
 export { applyCapabilities, CapabilityError } from './capabilities/index.js'
 export type { CapabilityContext } from './capabilities/index.js'
 export { materializeGit } from './capabilities/git.js'
--- a/packages/guard-runner/src/package-links.ts
+++ b/packages/guard-runner/src/package-links.ts
@@ -1,0 +1,176 @@
+/**
+ * Package-under-test resolution links. Scenario `setup.files` import the target
+ * package the way its docs tell users to — by published name (`import 'tsx'`,
+ * `require('tsx/cjs/api')`). The sandbox cwd is a bare temp dir, so without help
+ * Node rejects those bare specifiers (MODULE_NOT_FOUND) before any behavior of
+ * the package runs, and every programmatic-API scenario dies at birth.
+ *
+ * The engine therefore links each discovered package into the sandbox at
+ * `node_modules/<name>` → its real directory — `npm link` semantics. Node
+ * resolves through the link's realpath, so the package's own dependencies come
+ * from the repo's real `node_modules` and subpaths (`pkg/sub`) resolve through
+ * its actual `exports` map. Discovery covers the repo-root package and, for
+ * monorepos, every workspace package (npm/yarn `workspaces` globs and
+ * `pnpm-workspace.yaml`). A repo with no linkable package (non-Node targets)
+ * yields an empty list and the sandbox is unchanged.
+ */
+
+import fs from 'node:fs'
+import path from 'node:path'
+import yaml from 'js-yaml'
+
+export interface PackageLink {
+  /** The package's published name (`name` from its package.json). */
+  name: string
+  /** Absolute directory the sandbox `node_modules/<name>` link points at. */
+  dir: string
+}
+
+interface PackageJson {
+  name?: unknown
+  workspaces?: unknown
+}
+
+/**
+ * All packages a sandbox should make resolvable by name: the repo-root package
+ * plus every workspace package. Deduplicated by name (root first), sorted
+ * workspace order, packages without a valid name skipped.
+ */
+export function discoverPackageLinks(repoRoot: string): PackageLink[] {
+  const links: PackageLink[] = []
+  const seen = new Set<string>()
+  const add = (pkg: PackageJson | null, dir: string): void => {
+    const name = pkg?.name
+    if (typeof name !== 'string' || !isLinkableName(name) || seen.has(name)) return
+    seen.add(name)
+    links.push({ name, dir })
+  }
+  const rootPkg = readPackageJson(repoRoot)
+  add(rootPkg, repoRoot)
+  for (const dir of workspaceDirs(repoRoot, rootPkg)) add(readPackageJson(dir), dir)
+  return links
+}
+
+/**
+ * A name is linkable when `node_modules/<name>` stays inside `node_modules`: a
+ * bare name or a single-scope `@scope/name`, with no empty, `.`/`..`, or
+ * backslash segments. Anything else is an invalid npm name anyway.
+ */
+function isLinkableName(name: string): boolean {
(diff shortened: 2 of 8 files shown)


mushgev:
Closing unmerged by decision: guard doesn't support the programmatic-API (library) surface yet, and we're not investing in it now. #756 makes that explicit — extraction classifies import-by-name claims as a recorded-only `library` driver, so the #754 birth findings disappear at the source (those sections surface as honest "Needs library driver" gaps instead). The sandbox package-link mechanism built here (npm-link semantics, fully tested) is the execution floor to revive when the library driver ships — branch kept for that.

Issue #795, opened 2026-07-16
C# CLI e2e fixture is not restored before Roslyn project analysis


yahiaelmasry:
## Summary

The C# CLI end-to-end test copies tests/fixtures/sample-csharp-project-negative into a temporary directory and removes bin and obj, but it does not restore the copied solution before calling analyzeInProcess.

Roslyn project-aware analysis then rejects the Utils project with:

    project Utils loaded without a reference set

## Reproduction

Run the C# CLI end-to-end test:

    pnpm vitest run tests/cli/analyze-csharp.test.ts

The fixture setup in tests/cli/analyze-csharp.test.ts copies the project while excluding bin and obj, initializes Git, registers the project, and starts analysis without running dotnet restore.

## Root cause

Removing obj also removes NuGet restore metadata such as project.assets.json. Without a restore in the temporary copy, Roslyn cannot resolve framework and project references for the solution.

This is test-fixture setup rather than an EF schema parsing failure: the pipeline stops during project loading before the test can validate its EF assertions.

## Expected behavior

The copied fixture should be a valid restored C# solution before project-aware analysis begins, allowing the end-to-end test to exercise the complete C# pipeline.

## Suggested fix

Run dotnet restore in workDir after copyDir and before analyzeInProcess. If restore time can exceed the current setup timeout, update the beforeAll timeout accordingly.

yahiaelmasry:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Ensure every fresh C# CLI end-to-end fixture restores its own solution before project-aware Roslyn analysis.

**Current behavior:**
The dedicated C# CLI end-to-end case copies the C# fixture into a clean temporary repository without build or restore artifacts, then starts analysis before restoring the copied solution. An isolated run was confirmed to fail during project-aware analysis with `project 'Utils' loaded without a reference set — run dotnet restore`.

The generic multi-language CLI end-to-end suite has the same fresh-copy setup for its C# case and no explicit restore step. Source inspection shows that it can depend on restore artifacts left by another test, although that generic case was not runtime-reproduced during triage because its package build artifacts were unavailable in the verification checkout.

**Desired behavior:**
Every C# CLI end-to-end case that analyzes a fresh copied fixture must restore the copied solution before project-aware Roslyn analysis begins. Each case must own its setup completely: it must not rely on ignored restore artifacts in the source fixture, side effects from another test, or test execution order.

[...]

PR #796, opened 2026-07-16
fix(analyzer): reconcile EF Core schemas across files

### LINKED ISSUE #790: C# / EF Core: every table is emitted twice (entity-class name vs. `DbSet` name), which drops all relationships from the diagram
> **Note:** entity/column names below are anonymized placeholders (a generic content/publishing schema). The **counts, structure, and behavior are from a real run** and reproduce reliably.

## Environment

- **truecourse**: 0.7.2
- **OS**: macOS (Darwin)
- **Node**: v24.15.0
- **Analyzed project**: C# / .NET, EF Core (Npgsql/PostgreSQL), single-service (monolith), ~430 files
- **LLM provider**: claude-code (default)

## Summary

On a standard EF Core codebase, the C# schema parser produces **two tables for every entity** and never reconciles them:

- one from the **entity class name**, snake-cased (`Article` → `article`) — this copy has the real columns and **all** the foreign keys, and
- one from the `**DbSet<T>` property name** (`public DbSet<Article> Articles` → `Articles`) — this copy is an **empty stub** (0 columns, 0 FKs).

Because every relationship is attached to the *snake_case* copy but the two copies are never merged, the ER diagram cannot resolve most edges. In our project **28 of 42 relationships are undrawable**, and the service→database graph has **0 edges**. The database view looks essentially empty.

## What I expected

One table per entity (using the real table name), with foreign keys resolved to the correct target table, and the service connected to its database.

## What actually happened

The single PostgreSQL database came back with **47 “tables”** that break down as:

| Origin                                                                       | Count | Avg columns | FK columns       | Notes                              |
| ---------------------------------------------------------------------------- | ----- | ----------- | ---------------- | ---------------------------------- |
| Entity class name → snake_case (`article`, `revision`, `grant`, …)           | 21    | 11.3        | 42 (all of them) | the “real” data lives here         |
| `DbSet<T>` property name → PascalCase (`Articles`, `Revisions`, `Grants`, …) | 24    | 0           | 0                | empty stubs                        |
| Repository field `private readonly DbSet<T> _dbSet;`                         | 2     | 0           | 0                | junk node literally named `_dbSet` |

Downstream effects:

- `graph.databases[0].dbRelations` has 42 entries; **only 14 have both endpoints resolvable** to a node. 28 dangle.
- `graph.databaseConnections` is `[]` and `databases[0].connectedServices` is `[]` → the service node and database node are drawn with **no edge between them**.
- One entity (an activity-log table) was emitted under the name `**of**` (its columns are unmistakably an audit log: `timestamp, actor_id, entity_type, entity_id, success, …`) — a separate name mis-parse from the same class-reading path.

## Two root causes

**1. Two unreconciled schema readers.** The parser appears to collect tables from (a) entity class declarations and (b) `DbSet<T>` properties on the `DbContext`, but nothing maps `class Article` ⇄ `DbSet<Article> Articles` ⇄ `ToTable("Articles")` to a single table. So each entity is listed twice under different casing/pluralization, and the reader that finds columns/FKs is not the one that carries the canonical (real) table name.

- Real table names in our project are the PascalCase plurals — from `DbSet<Article> Articles` and migration `b.ToTable("Articles")`. Ironically these are the *empty* copies; all the detail is on snake_case names that don’t exist in the actual database.
- The `DbSet` reader is also too literal: it turned repository field declarations like `private readonly DbSet<Account> _dbSet;` into a table named `_dbSet`.

**2. FK targets are guessed by stripping `_id`.** Foreign keys are resolved to a table name by removing the `_id` suffix from the column, rather than following the navigation-property type or EF’s configured principal entity:

| FK column             | Resolved target                    | Actual target              |
| --------------------- | ---------------------------------- | -------------------------- |
| `owner_id`            | `owner` (no such table)            | `Accounts`                 |
| `parent_id`           | `parent` (no such table)           | self-ref → `Articles`      |
| `principal_id`        | `principal` (no such table)        | polymorphic (account/team) |
| `reviewer_id`         | `reviewer` (no such table)         | `Accounts`                 |
| `current_revision_id` | `current_revision` (no such table) | `Revisions`                |
| `source_layout_id`    | `source_layout` (no such table)    | `Layouts`                  |

23 of the 28 broken relations fail this way; the other 5 would resolve if the snake/Pascal duplication (cause 1) were fixed.

## Minimal repro (schema shape that triggers it)

```csharp
// Entities
public class Account { public Guid Id { get; set; } }

public class Article
{
    public Guid Id { get; set; }
    public Guid OwnerId { get; set; }      // FK
    public Account Owner { get; set; }     // navigation
    public Guid? ParentId { get; set; }    // self-ref FK
    public Article Parent { get; set; }
}

// DbContext
public class AppDbContext : DbContext
{
    public DbSet<Article> Articles { get; set; }
    public DbSet<Account> Accounts { get; set; }
}
```

**Expected:** 2 tables (`Articles`, `Accounts`) with edges `Articles.OwnerId → Accounts` and `Articles.ParentId → Articles`.

**Observed pattern (matches our full project):** duplicate `article`/`Articles` (and `account`/`Accounts`) nodes, the detailed copy under the snake_case name, and the relation drawn as `article → owner` (dangling) instead of `Articles → Accounts`.

## Impact

The database / ER view is unusable for C# EF Core projects: doubled nodes, empty stubs, junk `_dbSet` nodes, no service↔DB edge, and the large majority of relationships missing. Because this is deterministic (not the LLM tiers), re-running does not help.

## Suggested direction

- Canonicalize each entity to a single table by reconciling **class name ⇄ `DbSet<T>` property ⇄ `ToTable(...)` / `[Table]**` (and applied naming conventions) before building the graph; prefer the EF-configured table name.
- Ignore `DbSet<T>` declarations that are private fields (e.g. `_dbSet` in repositories); only the `DbContext`’s public `DbSet` properties define tables.
- Resolve FK targets via the navigation property’s type / EF principal entity, not by stripping `_id`; leave truly polymorphic or external FKs unlinked rather than inventing a target table.

Happy to test a fix or contribute a PR.

yahiaelmasry:
## Summary

- reconcile EF Core `DbContext`, entity, navigation, and fluent-configuration evidence across service files
- emit one canonical table per entity using `ToTable` > `[Table]` > `DbSet` > convention precedence
- resolve foreign keys through navigation types and canonical table identities instead of guessing from `*Id` names
- preserve model/provider scope across PostgreSQL, SQL Server, MySQL, and SQLite

## Root cause

The EF Core schema parser processed files independently. `DbContext` files emitted empty `DbSet` tables while entity files emitted separate detailed tables under class-derived names. Because foreign-key targets were guessed from scalar property names, many relation endpoints did not correspond to an emitted table.

## What changed

- added a service-scoped parser capability to the schema-parser registry
- introduced model reconciliation for split and partial entities/contexts, indirect `DbContext` inheritance, and explicitly applied `IEntityTypeConfiguration<T>` mappings
- retained canonical entity identity until tables, column references, and relation endpoints are emitted
- ignored private repository `DbSet` fields and unrelated convention-only POCOs
- associated `AddDbContext<T>(...UseX...)` provider evidence with the referenced context; ambiguous providers remain unresolved rather than defaulting
- masked comments, ordinary/verbatim/raw strings, and character literals before declaration extraction
- added focused parser and detector regression coverage for the reported failure modes

## Impact

Standard EF Core projects now produce detailed canonical tables and drawable relationships instead of duplicate entity/`DbSet` nodes and dangling guessed edges.

## Validation

- `vitest run tests/analyzer/efcore-parser.test.ts tests/analyzer/csharp-graph-snapshot.test.ts` — 29 passed
- `tsc -p packages/analyzer/tsconfig.json --noEmit`
- `tsc -p packages/analyzer/tsconfig.json`
- `dotnet build -c Release tools/csharp-roslyn-host` — 0 warnings, 0 errors
- `git diff --check`

The existing C# CLI e2e fixture has a separate restore/setup failure tracked in #795; this PR intentionally does not mix that test-infrastructure fix into the EF schema change.

Closes #790

### DIFF
--- a/packages/analyzer/src/database-detector.ts
+++ b/packages/analyzer/src/database-detector.ts
@@ -3,7 +3,8 @@ import { join, resolve } from 'path'
 import type { FileAnalysis, DatabaseType, DatabaseInfo, DatabaseConnectionInfo, DatabaseDetectionResult, TableInfo, RelationInfo } from '@truecourse/shared'
 import { DATABASE_IMPORT_MAP, DOCKER_IMAGE_MAP } from './patterns/database-patterns.js'
 import { parsePrismaSchema } from './schema-parsers/prisma.js'
-import { SCHEMA_PARSERS } from './schema-parsers/registry.js'
+import { parseEfCoreProject, type EfCoreProjectFile } from './schema-parsers/efcore.js'
+import { EF_CORE_SCHEMA_PARSER, SCHEMA_PARSERS } from './schema-parsers/registry.js'
 import { readAllDependencies } from './service-detectors/registry.js'
 import type { Service } from './service-detector.js'
 
@@ -123,9 +124,69 @@ export function detectDatabases(
     schemaResults.set(dbType, existing)
   }
 
+  // EF Core is a model-scoped ORM: DbContext declarations, entity classes,
+  // navigation properties, and fluent mappings normally live in separate
+  // files. Parse all matching files together per service so those facts can be
+  // reconciled before emitting tables and relations.
+  const efParser = EF_CORE_SCHEMA_PARSER
+  if (efParser.scope === 'service') {
+    const efFiles: EfCoreProjectFile[] = []
+    const serviceByFile = new Map<string, Service>()
+    for (const service of services) {
+      for (const filePath of service.files) serviceByFile.set(filePath, service)
+    }
+    const efServiceNames = new Set(
+      detections
+        .filter((detection) => detection.driver.startsWith('efcore-'))
+        .map((detection) => detection.serviceName),
+    )
+    for (const analysis of analyses) {
+      if (!efParser.matchesImport(analysis)) continue
+      const service = serviceByFile.get(analysis.filePath)
+      if (service) efServiceNames.add(service.name)
+    }
+
+    for (const analysis of analyses) {
+      if (analysis.language !== 'csharp') continue
+
+      const service = serviceByFile.get(analysis.filePath)
+      const serviceName = service?.name ?? '__root__'
+      const hasEfEvidence = efParser.matchesImport(analysis)
+      if (!hasEfEvidence && !efServiceNames.has(serviceName)) continue
+
+      try {
+        const content = readFileSync(resolve(analysis.filePath), 'utf-8')
+
+        const providerTypes = detections
+          .filter((detection) =>
+            detection.serviceName === serviceName && detection.driver.startsWith('efcore-')
+          )
+          .map((detection) => detection.type)
+
+        efFiles.push({
+          filePath: analysis.filePath,
+          content,
+          serviceName,
+          providerTypes,
+        })
+      } catch {
+        // Skip files that can't be read
+      }
+    }
+
+    for (const result of parseEfCoreProject(efFiles)) {
+      if (result.dbType === null) continue
+      const existing = schemaResults.get(result.dbType) || { tables: [], relations: [] }
+      existing.tables.push(...result.tables)
+      existing.relations.push(...result.relations)
+      schemaResults.set(result.dbType, existing)
+    }
+  }
+
   // Import-based schema parsers (Drizzle, SQLAlchemy, etc.)
   for (const analysis of analyses) {
     for (const parser of SCHEMA_PARSERS) {
+      if (parser.scope === 'service') continue // handled repository-wide above
       if (!parser.matchesImport(analysis)) continue
 
       try {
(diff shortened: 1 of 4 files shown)


DoilLee2222:
Reviewed this closely — the cross-file reconciliation is solid and directly fixes the duplication in #790. Two gaps still left standard EF diagrams incomplete, so I opened #800 which builds on this commit (cherry-picked, attribution preserved) and adds:

1. **Collection-only 1:N relations** — FK resolution required a reference nav on the dependent side, so the common `Order.Lines` (collection on principal) + bare `OrderLine.OrderId` shape stayed unlinked. #800 resolves `<X>Id` to a modeled, bound entity `X` (EF's `<PrincipalType>Id` convention), still leaving `owner_id`-with-no-`Owner` unlinked.
2. **Enum columns** — enum-typed properties were classified as navigations and dropped; #800 keeps them as scalar columns.

Full details and before/after in #800. Thanks for the thorough work here 🙏

---

spiderhands:
Merged in #800

PR #798, opened 2026-07-16
test(cli): restore C# fixtures before Roslyn analysis

### LINKED ISSUE #795: C# CLI e2e fixture is not restored before Roslyn project analysis

yahiaelmasry:
## Summary

- restore each temporary C# CLI fixture before project-aware Roslyn analysis
- prevent the generic C# fixture copy from inheriting stale `bin` or `obj` artifacts
- allow C# fixture setup enough time for a real solution restore

## Root cause

The CLI end-to-end tests copied C# solutions into temporary repositories and analyzed them without restoring the copied projects. Roslyn correctly rejected the clean projects because their framework reference sets were unavailable. The generic suite could appear to pass when another test had already restored the shared source fixture and left ignored build artifacts behind.

Each C# case now owns its setup: it excludes generated artifacts, restores only its temporary solution, and starts analysis after restore succeeds. Production Roslyn behavior is unchanged.

## Validation

- full workspace build: 19/19 tasks passed
- Roslyn host Release build: succeeded with 0 errors
- dedicated C# CLI end-to-end test: passed independently
- generic C# CLI end-to-end case: passed independently
- relevant CLI suite: 9/9 tests passed
- shared C# source fixture remained free of generated `bin` and `obj` directories
- `git diff --check`

Closes #795

### DIFF
--- a/tests/cli/analyze-csharp.test.ts
+++ b/tests/cli/analyze-csharp.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { execSync } from 'node:child_process';
+import { execFileSync, execSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 
 // Same socket stub as analyze.test.ts — emits require a live socket server.
@@ -67,11 +67,16 @@ describe('CLI analyze pipeline (e2e, C#)', () => {
     execSync('git init -q -b main', { cwd: workDir, env });
     execSync('git add -A', { cwd: workDir, env });
     execSync('git -c commit.gpgsign=false commit -q -m init', { cwd: workDir, env });
+    execFileSync('dotnet', ['restore', 'SampleCsharpProject.sln'], {
+      cwd: workDir,
+      env,
+      encoding: 'utf8',
+    });
 
     project = await registerProject(workDir);
     await updateProjectConfig(workDir, { enableLlmRules: false });
     clearLatestCache();
-  }, 30_000);
+  }, 300_000);
 
   afterAll(async () => {
     if (project) await unregisterProject(project.slug);
--- a/tests/cli/analyze.test.ts
+++ b/tests/cli/analyze.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { execSync } from 'node:child_process';
+import { execFileSync, execSync } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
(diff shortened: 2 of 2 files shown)


DoilLee2222:
Reviewed the diff — the approach is correct and well-scoped. Thanks for the clean fix. ✅

**What it gets right**
- Both copied-fixture e2e cases (`analyze-csharp.test.ts` and the C# case in `analyze.test.ts`) now restore *their own* temporary solution before project-aware analysis, so neither depends on another test's restore artifacts or on execution order.
- The generic suite is handled correctly per-language: restore + `bin`/`obj` exclusion are gated on `c.lang === 'C#'`, and the `skipDotnetBuildArtifacts` flag is threaded through the recursion.
- Restore runs *after* the git commit, so the regenerated `obj/` stays untracked (and it's ignored by analysis via the C# `**/obj/` pattern anyway) — no build artifacts get committed.
- Using `execFileSync(..., { encoding: 'utf8' })` (rather than `stdio: 'ignore'`) means a failed restore throws with its diagnostic captured, so setup fails hard with a real error instead of letting analysis proceed on an unrestored project.
- Timeout bump to 300s matches the existing project-mode restore convention.

**CI status (not a code issue)**
- The `Tests` workflow is currently `action_required` — this is the standard gate for a first-time-contributor fork PR and needs a maintainer to click **Approve and run** before the real suite (incl. the C# e2e tests with the Roslyn host + `dotnet restore`) executes. That green run is the check that actually matters here.
- The failing **Vercel** check is `Authorization required to deploy` — a deploy-preview auth gate for fork PRs, unrelated to correctness. Since this PR only touches `tests/cli/*.ts` (nothing under `apps/landing` or the dashboard), the preview is irrelevant and can be ignored for merge.

Once the approved `Tests` run is green, this looks good to merge.

---

Issue #791, opened 2026-07-15
`analyze`: no resume / partial re-run — a rate-limit near the end forces a full ~2.8M-token re-run


yahiaelmasry:
## Environment

- **truecourse**: 0.7.2
- **OS**: macOS (Darwin)
- **Node**: v24.15.0
- **LLM provider**: claude-code (default), `maxConcurrency: 10`
- **Project**: ~432 files; pre-flight estimate **2,805,898 tokens across 20 tiers**

## Summary

`truecourse analyze` runs every LLM tier in one pass. If the provider hits a **session rate limit (HTTP 429)** partway through — which is easy on a multi-million-token run — the tiers that hadn’t finished simply fail, yet the analysis is still saved as `status: "completed"`. There is no way to resume or to re-run only the failed tiers, so recovering the missing pieces means paying the **entire ~2.8M-token cost again**.

## What happened

Near the end of an otherwise-successful run, the claude-code provider began returning:

```
api_error_status: 429  "You've hit your session limit · resets 6:40pm"
```

The tiers still in flight at that moment all failed:

```
[WARN] [LLM] database-schema: failed in 1336287ms — claude exited with code 1: ... 429
[INFO] [CLI Violations] service call failed: ... 429
[INFO] [CLI Violations] module call failed: ... 429
```

Two issues compounded it:

1. **~22 minutes were spent retrying against a hard session limit** (`failed in 1336287ms`). Retries can’t succeed until the session window resets, so the backoff should account for the reset time rather than burning attempts.
2. The run was still recorded as `status: "completed"`, and the dashboard shows the partial result (e.g. “Database checks — Schema LLM failed”) with **no affordance to retry just that tier**. The only documented recovery is `truecourse analyze` again — a full re-run.

## Impact

- A transient rate-limit in the last few percent of a run discards that work and forces a full, expensive re-analysis.
- Deterministic results were persisted fine; only a handful of LLM calls failed — but there’s no way to target just them.

## Suggestions (any one of these would help)

- **Checkpoint + resume**: persist per-tier/per-batch completion; on the next `analyze`, skip tiers already completed for the same commit and only run what’s missing. A `--resume` flag would be ideal.
- **Targeted re-run**: `--only <domain>` (e.g. `--only database`) and/or `--retry-failed` to re-run just the tiers marked failed in the last analysis.
[...]

PR #800, opened 2026-07-16
fix(analyzer): reconcile EF Core schemas across files (+ collection-only relations & enum columns)

### LINKED ISSUE #790: C# / EF Core: every table is emitted twice (entity-class name vs. `DbSet` name), which drops all relationships from the diagram

DoilLee2222:
## Summary

Fixes the EF Core schema duplication reported in #790, and closes two remaining gaps that still left standard EF diagrams incomplete.

This branch builds on @yahiaelmasry's #796 (cherry-picked, attribution preserved) and adds two follow-up fixes on top.

## Included from #796

- Reconcile `DbContext`, entity, navigation, and fluent-configuration evidence **across files** → one canonical table per entity (`ToTable` > `[Table]` > `DbSet` > convention).
- Resolve foreign keys through navigation types instead of guessing from `*Id` names; ignore private repository `DbSet` fields; associate `AddDbContext<T>(UseX)` providers per context.

## Added here

1. **Collection-only 1:N relations are now drawn.** FK resolution previously required a reference navigation on the dependent side, so the common parent-child shape — the principal owns the only navigation as a collection (`Order.Lines`), the dependent carries a bare `OrderId` — was left unlinked. We now resolve `<X>Id` to a modeled, **bound** entity named `X` (EF's `<PrincipalType>Id` convention). Restricted to real entities, so unrelated scalars like `owner_id` with no `Owner` entity stay unlinked rather than guessed.
2. **Enum columns are kept.** Enum-typed properties (PascalCase, non-BCL) were classified as navigations and dropped. We collect `enum` declarations across the service and reclassify those properties as scalar columns.

## Behavior, before → after (split context/entity, enum col, collection nav)

```
before  Orders: [id, customer_id, total_cents]                 relations: []
after   Orders: [id, customer_id, status, total_cents]         relations: [OrderLines→Orders (order_id)]
```

`customer_id` (no `Customer` entity) correctly stays unlinked — no invented endpoint.

## Validation

- `efcore-parser.test.ts` — 26 passed (23 from #796 + 3 new: collection-only edge, no-invented-target safety, enum-column presence)
- `csharp-graph-snapshot.test.ts` — 6 passed (no regression)
- full `tests/analyzer` suite green; `tsc --noEmit` and analyzer build clean

## Scope

EF schema only. The C# CLI e2e restore (#795) and analyze resume (#791) are tracked separately and intentionally untouched here.

Closes #790

---
_Generated by [Claude Code](https://claude.ai/code/session_0126DEjHGe2CKXv1eE2PVT9b)_

### DIFF
--- a/packages/analyzer/src/database-detector.ts
+++ b/packages/analyzer/src/database-detector.ts
@@ -183,12 +183,27 @@ export function detectDatabases(
       }
     }
 
+    // Last-resort provider for a model the parser could not type (no inline
+    // OnConfiguring, no AddDbContext hint, and its own scope had no provider
+    // package — e.g. a DbContext in a class library outside every service).
+    // Only fall back when the whole repo names exactly one EF provider, so an
+    // ambiguous multi-provider repo still drops the untyped schema.
+    const efProviderTypes = [
+      ...new Set(
+        detections
+          .filter((detection) => detection.driver.startsWith('efcore-'))
+          .map((detection) => detection.type),
+      ),
+    ]
+
     for (const result of parseEfCoreProject(efFiles)) {
-      if (result.dbType === null) continue
-      const existing = schemaResults.get(result.dbType) || { tables: [], relations: [] }
+      const dbType =
+        result.dbType ?? (efProviderTypes.length === 1 ? efProviderTypes[0]! : null)
+      if (dbType === null) continue
+      const existing = schemaResults.get(dbType) || { tables: [], relations: [] }
       existing.tables.push(...result.tables)
       existing.relations.push(...result.relations)
-      schemaResults.set(result.dbType, existing)
+      schemaResults.set(dbType, existing)
     }
   }
 
--- a/packages/analyzer/src/database-detector.ts
+++ b/packages/analyzer/src/database-detector.ts
@@ -15,6 +15,9 @@ interface DetectedDatabase {
   connectionEnvVar?: string
 }
 
+/** Model scope for C# files that belong to no detected service. */
+const ROOT_MODEL_SCOPE = '__root__'
+
 /**
  * Detect databases used in the codebase by scanning imports, schema files,
  * and Docker Compose configuration.
@@ -135,22 +138,28 @@ export function detectDatabases(
     for (const service of services) {
       for (const filePath of service.files) serviceByFile.set(filePath, service)
     }
+    // The model scope of a C# file: its detected service, or the shared root
+    // bucket for files owned by none. EF entities and their DbContext commonly
+    // live in class libraries (ApplicationCore/Infrastructure) that are not
+    // detected services, so both fall here and must reconcile as one model.
+    // Both loops below derive scope through this single helper so a file can
+    // never register under one name yet be collected under another.
+    const serviceNameOf = (analysis: FileAnalysis): string =>
+      serviceByFile.get(analysis.filePath)?.name ?? ROOT_MODEL_SCOPE
+
     const efServiceNames = new Set(
       detections
         .filter((detection) => detection.driver.startsWith('efcore-'))
         .map((detection) => detection.serviceName),
     )
     for (const analysis of analyses) {
-      if (!efParser.matchesImport(analysis)) continue
-      const service = serviceByFile.get(analysis.filePath)
-      if (service) efServiceNames.add(service.name)
+      if (efParser.matchesImport(analysis)) efServiceNames.add(serviceNameOf(analysis))
     }
 
     for (const analysis of analyses) {
       if (analysis.language !== 'csharp') continue
 
-      const service = serviceByFile.get(analysis.filePath)
-      const serviceName = service?.name ?? '__root__'
+      const serviceName = serviceNameOf(analysis)
       const hasEfEvidence = efParser.matchesImport(analysis)
       if (!hasEfEvidence && !efServiceNames.has(serviceName)) continue
 
--- a/packages/analyzer/src/database-detector.ts
+++ b/packages/analyzer/src/database-detector.ts
@@ -3,7 +3,8 @@ import { join, resolve } from 'path'
 import type { FileAnalysis, DatabaseType, DatabaseInfo, DatabaseConnectionInfo, DatabaseDetectionResult, TableInfo, RelationInfo } from '@truecourse/shared'
 import { DATABASE_IMPORT_MAP, DOCKER_IMAGE_MAP } from './patterns/database-patterns.js'
 import { parsePrismaSchema } from './schema-parsers/prisma.js'
-import { SCHEMA_PARSERS } from './schema-parsers/registry.js'
+import { parseEfCoreProject, type EfCoreProjectFile } from './schema-parsers/efcore.js'
+import { EF_CORE_SCHEMA_PARSER, SCHEMA_PARSERS } from './schema-parsers/registry.js'
 import { readAllDependencies } from './service-detectors/registry.js'
 import type { Service } from './service-detector.js'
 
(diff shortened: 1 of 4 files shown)


spiderhands (changes requested):
Overall this is in good shape — the cross-file reconciliation works, the two additions on top of #796 (collection-only 1:N relations, enum columns) are correct and well-tested, and attribution to #796 is preserved properly. One blocking issue before merge:

**C# 12 primary-constructor DbContexts produce an empty schema.**

The .NET 8+ template default shape is not recognized as a context:

```csharp
public class AppDbContext(DbContextOptions<AppDbContext> options) : DbContext(options)
{
    public DbSet<Article> Articles { get; set; }
}
```

`extractBaseTypes` yields the base entry `DbContext(options)`, and `simpleTypeName()` strips generic arguments but not the constructor argument list, so the `=== 'DbContext'` check in `parseProjectFile` never matches. The class falls through as an entity, no context is found, and `parseEfCoreProject` returns `[]` — zero tables. This is a regression vs `main`, where the old parser at least emitted a name-only `Articles` stub table for the same input.

Suggested fix: strip a trailing `(...)` argument list when normalizing base-type references (e.g. in `normalizeNamedTypeReference`), which also covers the indirect-inheritance path (`: BaseContext(options)`) since it uses the same normalization.

Please add this regression test to `tests/analyzer/efcore-parser.test.ts` — it fails on the current branch:

```ts
it('recognizes a primary-constructor DbContext (.NET 8 template style)', () => {
  const result = parseEfCoreSchema(`
    using Microsoft.EntityFrameworkCore;
    namespace Publishing;

    public class AppDbContext(DbContextOptions<AppDbContext> options) : DbContext(options)
    {
      public DbSet<Article> Articles { get; set; }
    }

    public class Article { public Guid Id { get; set; } }
  `)

  expect(result.tables.map((t) => t.name)).toEqual(['Articles'])
})
```

One housekeeping ask: please remove the `Generated by Claude Code (claude.ai/code/session…)` line from the PR description — session links don't belong in PR bodies per repo policy.

spiderhands:
Tested this branch against [eShopOnWeb](https://github.com/dotnet-architecture/eShopOnWeb) (commit `d46afba`) and the reconciliation doesn't take effect there — every table still comes back with 0 columns and no relations are drawn.

The cause is upstream of this PR, in `database-detector.ts`: files that belong to no detected service land in a `'__root__'` bucket that never registers as an EF service, so they're filtered out before the parser runs. eShopOnWeb keeps its entity POCOs in an `ApplicationCore` class library with the EF configuration in a separate `Infrastructure` project — neither is a detected service, and the POCOs import no EF namespace of their own, so they get dropped. Those POCOs are what carry the properties that become columns.

So the parser work in this PR looks correct; the property-bearing files just never reach it. Letting the `'__root__'` bucket register as an EF service recovers all seven tables' columns plus three FK relations — including the collection-only case this PR adds.

Worth noting this isn't specific to eShopOnWeb: entities in a domain library with EF config in a separate infrastructure project is standard Clean Architecture layout, so any repo following it would hit the same gate.

packages/analyzer/src/database-detector.ts
```diff
@@ -123,9 +124,69 @@ export function detectDatabases(
...
+    }
+    const efServiceNames = new Set(
+      detections
+        .filter((detection) => detection.driver.startsWith('efcore-'))
+        .map((detection) => detection.serviceName),
+    )
+    for (const analysis of analyses) {
+      if (!efParser.matchesImport(analysis)) continue
+      const service = serviceByFile.get(analysis.filePath)
+      if (service) efServiceNames.add(service.name)
```
spiderhands: Files that belong to no detected service fall into the `'__root__'` bucket below, but this `if (service)` guard means `'__root__'` never registers as an EF service — so the gate on line 152 filters those files out before the parser sees them.

This bites on Clean Architecture layouts where entity POCOs live in a domain library that isn't a detected service and imports no EF namespace of its own. Those POCOs carry the properties that become columns, so tables come back column-less.

```suggestion
      efServiceNames.add(service?.name ?? '__root__')
```

DoilLee2222:
Great catch — fixed in `b5e61ae`.

Root cause was exactly as you described: a file's model scope was computed two ways that disagreed. The collection loop mapped service-less files to `__root__`, but the registration loop skipped files with no service, so `__root__` never became an EF service and the property-bearing POCOs were dropped before the parser ran. Both loops now derive scope through one helper, so files owned by no detected service reconcile as one root-scoped model.

Verified on a Clean Architecture layout (entities in `ApplicationCore`, context in `Infrastructure`, provider pkg in `Web`): recovers the columns and the FK relation instead of empty stubs. Added a detector regression test for it. Thanks for testing against eShopOnWeb 🙏

---

spiderhands (approved):

PR #815, opened 2026-07-17
fix(llm): stop analysis on Claude session limits

yahiaelmasry:
## Summary

- recognize definite Claude session-limit responses across direct CLI, installed, streaming, and agent-mailbox transports while preserving retries and diagnostics for generic 429 responses
- open a provider-wide circuit on the first definite limit so queued calls and retries from already-admitted peers do not make unnecessary external requests
- surface the reset hint immediately through the existing analysis tracker, including that queued work stopped while active calls finish
- propagate the terminal failure through LLM aggregation so the interrupted run does not update `LATEST.json`, history, or the canonical analysis chain

## Root cause

Claude reports an exhausted session with an error result whose API status is 429. Depending on the transport, that result can arrive as a native error envelope or the installed CLI wrapper `claude API error (api 429): ...`. The provider previously treated the wrapped response like every other process/parse failure and retried it, while streaming and agent-mailbox adapters could replace or generically wrap the native envelope before it reached the provider.

Concurrent calls were already queued behind the provider limiter, and several `Promise.allSettled()` aggregation points logged and omitted rejected LLM work. That combination could keep spending attempts after the session was unavailable and then persist an incomplete run as completed.

This change matches the structural error fields or the exact wrapped Claude API-429 shape together with the session-limit phrase, extracts the variable reset hint, and leaves a bare/generic 429 on the existing bounded-retry path. Cancellation remains a separate, higher-priority path.

## Scope and tradeoff

This is the first damage-containment slice for #791, not the complete fix, and it must not close the issue.

The interrupted run is deliberately not written into `LATEST.json` or the completed-analysis chain. Successful LLM calls from that interrupted run are not reusable yet, so rerunning can spend those tokens again. This PR limits additional waste after a definite session-limit response and protects the last completed baseline; it does not claim token preservation or resumability.

The complete design distinguishes the **latest attempted run** from the **latest completed analysis**. This PR protects only the latter: a blocked attempt cannot replace the trustworthy completed findings baseline. Later slices will store the attempted run in a separate durable journal with explicit `running`, `blocked`, `failed`, `finalizing`, and `completed` states; expose completed/pending/failed counts, reset time, and Resume through the CLI/dashboard/API; and atomically promote that run to the completed baseline only after its full certified plan succeeds. It will not be implemented by adding `status: "incomplete"` to the current canonical `LATEST.json`; an envelope redesign would require a migration of every reader.

Checkpoints are not being rushed into this slice because current LLM outputs can contain analysis graph IDs that change between runs. Saving and replaying those resolved outputs without stable work identity and semantic fingerprints could attach findings to the wrong nodes or falsely resolve findings.

Required follow-up contributions in the same #791 plan will:

- assign every LLM work unit a stable identity and sufficient input fingerprint
- durably store each successful result in a separate run journal outside `LATEST.json` and the canonical chain
- validate repository content, baseline, rules, configuration, prompts/schemas, provider, and model before reuse
- resume only pending or failed work and preserve usage accounting without double-counting reused results
- finalize `LATEST.json` only after the complete work plan succeeds, or report mixed success/failure and explicitly ask whether to bypass the integrity gate and save anyway
- allow interrupted runs to resume and/or restart their interrupted provider sessions after the reported limit reset

## Validation

- focused provider, shared/agent transport, and production analysis-integrity matrix — 5 files and 52/52 tests passed
- `dotnet build -c Release tools/csharp-roslyn-host` — succeeded with 0 errors; two NU1900 warnings because the NuGet vulnerability feed was unavailable
- `git diff --check`
- fresh independent spec and standards audits — no P0-P3 findings

The focused tests cover the observed Claude response variants, native and exact installed-wrapper forms, exit-code 0 and nonzero streaming envelopes, structured and nullable mailbox responses, generic-429 retries and existing generic diagnostics, queued and already-admitted circuit behavior, cancellation precedence, circuit reset/one-shot notification, visible reset communication, and isolated service/code/database failures with unchanged canonical files.

One full-suite run was attempted after the required builds but did not pass in this restricted environment. Its visible failures were dominated by sandbox-denied socket listeners and home-directory writes, fixture URL paths under a checkout path containing spaces, unavailable experimental Node `localStorage`, the known unrestored .NET fixture, and production-test hook timeouts under the large concurrent run. The same production integrity file passed 3/3 in isolation and in the combined focused matrix above. CI remains the authoritative broad-suite run.

Part of #791

### DIFF
--- a/packages/core/src/commands/analyze-in-process.ts
+++ b/packages/core/src/commands/analyze-in-process.ts
@@ -15,6 +15,10 @@ import { persistFullAnalysis, type PersistFullResult } from './analyze-persist.j
 import { config } from '../config/index.js';
 import { log } from '../lib/logger.js';
 import {
+  isLlmSessionLimitError,
+  LlmSessionLimitError,
+} from '@truecourse/shared/llm';
+import {
   bucketDuration,
   bucketFileCount,
   detectLanguages,
@@ -60,6 +64,14 @@ export interface AnalyzeInProcessOptions {
 
 export type AnalyzeInProcessResult = PersistFullResult;
 
+export class AnalysisSessionLimitError extends LlmSessionLimitError {
+  constructor(error: LlmSessionLimitError) {
+    super(error.resetHint);
+    this.name = 'AnalysisSessionLimitError';
+    this.message = `${this.message} The interrupted run was not saved, so LATEST.json was not updated and any previous completed analysis remains unchanged. Successful LLM calls from this interrupted run cannot be resumed yet and may be repeated when you rerun.`;
+  }
+}
+
 export async function analyzeInProcess(
   project: RegistryEntry,
   options: AnalyzeInProcessOptions = {},
@@ -68,7 +80,13 @@ export async function analyzeInProcess(
   log.info(
     `[LLM] Provider: claude-code, model: ${config.claudeCodeModel || 'default'}, maxConcurrency: ${config.claudeCodeMaxConcurrency}`,
   );
-  const core = await analyzeCore(project, { ...options, mode: 'full' });
+  let core: Awaited<ReturnType<typeof analyzeCore>>;
+  try {
+    core = await analyzeCore(project, { ...options, mode: 'full' });
+  } catch (error) {
+    if (isLlmSessionLimitError(error)) throw new AnalysisSessionLimitError(error);
+    throw error;
+  }
   const result = await persistFullAnalysis(project, core, startedAt);
 
   if (options.source) {
--- a/packages/core/src/services/llm/cli-provider.ts
+++ b/packages/core/src/services/llm/cli-provider.ts
@@ -10,7 +10,12 @@ import { registerChildProcess, unregisterChildProcess } from '../analysis-regist
 import { zodToJsonSchema } from 'zod-to-json-schema';
 import type { ZodType } from 'zod';
 import type { Violation } from '@truecourse/shared';
-import type { LlmTransport } from '@truecourse/shared/llm';
+import {
+  isLlmSessionLimitError,
+  parseLlmSessionLimitError,
+  type LlmSessionLimitError,
+  type LlmTransport,
+} from '@truecourse/shared/llm';
 import { config } from '../../config/index.js';
 import {
   getPrompt,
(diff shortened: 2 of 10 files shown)


DoilLee2222 (commented):
## Independent verification ✅

I checked out this PR's head (`5702c28`) into a clean worktree, installed with a frozen lockfile, and ran the build, lint, the new session-limit tests, and the closest existing regression tests. Everything I ran passed.

### Results

| Check | Command | Result |
| --- | --- | --- |
| Build | `pnpm build` | ✅ 19/19 tasks |
| Lint | `pnpm lint` | ✅ 14/14 tasks |
| New session-limit tests (4 files) | `vitest run` | ✅ 50/50 |
| Related regression tests (4 files) | `vitest run` | ✅ 25/25 |

**Total: 75 tests, all green**, on top of a clean build and lint. This matches the build (19/19) and lint (14/14) numbers reported in the PR description.

### New tests exercised

- `tests/shared/llm-transport.test.ts` — 37 tests
- `tests/core/cli-provider-session-limit-circuit.test.ts` — 6 tests
- `tests/core/cli-provider-session-limit.test.ts` — 4 tests
- `tests/core/analyze-session-limit-integrity.test.ts` — 3 tests

The behaviors that matter for this change are all directly asserted and passing:

- **Definite vs. generic 429 are distinguished.** A wrapped `claude API error (api 429): You've hit your session limit …` stops after a single attempt, while a generic `429 Rate limited` still follows the bounded-retry path (3 attempts). No over-broad matching.
- **Provider-wide circuit works.** Once one admitted call hits the limit, queued calls are not started and already-admitted peers do not retry after the circuit opens.
- **Baseline integrity holds.** When the limit is hit at `analyze.service`, `analyze.code`, or `analyze.database`, `LATEST.json` and `history.json` are byte-for-byte unchanged (verified via sha256 in the test) and the terminal error carries the `LLM_SESSION_LIMIT` code plus the reset hint.
- **Cancellation stays separate** and takes precedence over the circuit; **the circuit resets** on a new `setAnalysisId`.

### Regression tests exercised (no breakage)

`tests/core/cli-provider.test.ts`, `cli-provider-concurrency.test.ts`, `llm-agent-transport.test.ts`, `violation-lifecycle.test.ts` — 25/25 passing, so the transport/provider refactor did not disturb existing behavior.

### Scope note

I did **not** run the full suite. This repo's complete suite has environment-dependent failures unrelated to this PR (sandboxed socket listeners, the build-required C# Roslyn host, etc.) — the same limitation already documented under *Validation*. I deliberately scoped verification to the changed code and its immediate neighbors; CI remains the authoritative broad-suite run.

### Conclusion

Within the verified scope, the change does exactly what it claims: it recognizes a definite Claude session limit across transports, opens the circuit to stop wasted calls, surfaces the reset hint, and protects the last completed baseline from being overwritten by an interrupted run. The PR is also appropriately honest that this is the first damage-containment slice of #791 — it stops the bleeding (retries + baseline corruption) but does **not** yet add resumability or token preservation, which correctly remain follow-up work. No blocking issues found.


---

PR #816, opened 2026-07-18
feat(llm): retain exact context batch provenance

yahiaelmasry:
## Summary

- add structured source provenance to analyze LLM context batches
- retain requested metadata fields, exact targeted function ranges, and full-file membership
- keep provenance, content, file paths, and targeted function counts aligned through oversized file-boundary splits
- preserve the existing file and prompt order

## Why

The resumability work in #791 needs each future LLM work item to describe exactly which repository source it represents. Metadata and targeted batches currently carry only rendered text, while full-file batches carry paths separately. That is not a safe foundation for stable work identity or result ownership.

This focused slice records that scope structurally at the router boundary without changing execution or persistence.

## Scope boundary

This is an in-memory prerequisite only. The new `sources` field is not consumed by execution yet.

It does **not** add stable work IDs, fingerprints, repository-relative path normalization, deterministic routing, a durable journal, replay, token reuse, resume behavior, usage reconciliation, or any `LATEST.json`/canonical-analysis change. Successful calls from an interrupted run are still not reusable, so rerunning can still spend those tokens again.

The later #791 contributions still must add the provider identity seam, complete work manifest, exact lifecycle ownership, database and architecture failure ownership, durable per-work journaling, compatibility validation, pending/failed-only resume, exact-once usage, reset-aware continuation, and guarded finalization.

Part of #791. This PR does not resolve or close it.

## Validation

- `pnpm exec vitest run tests/core/llm-context-scope.test.ts` — 5 tests passed
- `dotnet build -c Release tools/csharp-roslyn-host` — 0 errors; two NU1900 vulnerability-feed warnings
- fresh independent spec, correctness, and test audits — no remaining P1/P2 findings

The focused tests cover included and excluded metadata/targeted/full-file scope, plus full-file and targeted oversized-split alignment. GitHub Actions remains the authoritative broad-suite run; the restricted local broad run was not used as passing evidence because its socket-dependent route tests cannot listen in this environment.

### DIFF
--- a/packages/core/src/services/llm/context-router.ts
+++ b/packages/core/src/services/llm/context-router.ts
@@ -14,6 +14,23 @@ export interface ContextBatch {
   estimatedTokens: number;
   /** Real file paths included in this batch (full-file/legacy tiers only). */
   filePaths?: string[];
+  /** Exact repository source scope represented by this batch. */
+  sources: ContextBatchSource[];
+}
+
+export interface ContextBatchSource {
+  path: string;
+  selection:
+    | { kind: 'metadata'; fields: MetadataField[] }
+    | { kind: 'full-file' }
+    | {
+        kind: 'targeted';
+        functions: Array<{
+          name: string;
+          startLine: number;
+          endLine: number;
+        }>;
+      };
 }
 
 export interface PreFlightEstimate {
@@ -186,7 +203,7 @@ function extractTargetedFunctions(
 // Metadata summary builder
 // ---------------------------------------------------------------------------
 
-type MetadataField = NonNullable<ContextRequirement['metadataFields']>[number];
+export type MetadataField = NonNullable<ContextRequirement['metadataFields']>[number];
 
 function buildMetadataSummary(fa: FileAnalysis, fields: MetadataField[]): string {
   const parts: string[] = [`=== ${fa.filePath} ===`];
@@ -312,7 +329,7 @@ function buildMetadataContent(
   group: { rules: RuleDto[]; requirement: ContextRequirement },
   fileAnalyses: FileAnalysis[],
   fileContents: Map<string, { content: string; lineCount: number }>,
-): { content: string; fileCount: number } {
+): { content: string; fileCount: number; sources: ContextBatchSource[] } {
   const matching = fileAnalyses.filter((fa) => {
     if (!group.requirement.fileFilter) return true;
     const fc = fileContents.get(fa.filePath);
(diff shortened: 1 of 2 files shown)


DoilLee2222:
## Review + verification summary

**Verdict: ✅ Looks good — verified locally, no correctness issues found.** All stated goals hold, the core invariant is sound, and the change is a safe, well-scoped in-memory prerequisite. A couple of non-blocking robustness suggestions are noted at the end.

---

### 1. Local verification

Checked out the PR head (`ce41614`) and reproduced the validation independently:

| Check | Command | Result |
|-------|---------|--------|
| PR unit tests | `pnpm exec vitest run tests/core/llm-context-scope.test.ts` | ✅ 5/5 passed |
| Build | `pnpm build` | ✅ 19/19 tasks |
| Lint | `pnpm lint` | ✅ 14/14 tasks |
| Regression sweep | `pnpm exec vitest run tests/core` | ✅ 39 files / 321 tests passed |

No regressions in the surrounding core suite. The full broad suite was **not** used as local evidence — it needs listening sockets (route tests) and the Roslyn host — so GitHub Actions remains authoritative there. The one red status on the PR is `Vercel — "Authorization required to deploy"`, which is a deploy-authorization gate, not a code/test failure.

### 2. Scope claim confirmed

The PR states the new `sources` field is produced but **not consumed by execution yet**. Verified: the only consumer of `routeContext()` is `violation-pipeline.service.ts`, and it reads `batch.filePaths` only — never `batch.sources`. `ContextBatch` is constructed exclusively inside `context-router.ts`, and the batch object is not persisted (`analysis-persistence.service.ts` serializes a separate `LlmDetail.filePaths`, not `ContextBatch`). So this is a genuinely inert, in-memory addition — low risk to merge.

### 3. Correctness — the key invariant

The subtle part is oversized-batch splitting: `splitIntoBatches` re-associates `sources[si]` with content sections positionally after `content.split(/(?=^=== )/m)`. This is only correct if `=== ` at line-start marks **exactly** the file boundaries and nothing else. I checked all three tiers:

- **full-file / targeted** — body lines are line-number-prefixed (`${n}: ...`), so a source line beginning with `=== ` renders as `12: === ...` and cannot match the split lookahead.
- **metadata** — field lines are label-prefixed (`Functions:`, `Imports:`, …) and comma-joined (no embedded newlines), so no body line starts with `=== `.

⟹ Sections align 1:1 with `parts`/`filePaths`/`sources` by index across every tier. The split-alignment logic is sound, and the two oversized-split tests (`full-file` and `targeted`) lock this behavior in. 👍

### 4. Nice latent fix worth calling out

In the split path, targeted batches now carry an accurate **per-batch** `functionCount` via `targetedFunctionCount(currentSources)`. Previously split targeted batches were pushed without `functionCount` (i.e. `undefined`), so this quietly corrects the count for large targeted groups in addition to adding provenance.

### 5. Non-blocking suggestions

These are hardening ideas for the resumability foundation this PR is meant to be — not merge blockers:

1. **Make `sources` a required parameter of `splitIntoBatches`** (drop the `= []` default). All three current callers pass it; the default only exists to satisfy the signature. Since provenance integrity is the whole point, a future fourth caller that forgets the argument would silently emit empty provenance instead of failing at compile time.
2. **Consider pairing sections with their source objects structurally** rather than re-joining and re-splitting text by index. Correctness currently rests on the `=== ` delimiter never appearing at the start of a body line — true today thanks to the numbering/label conventions above, but it couples provenance integrity to the rendered text format. Carrying `[section, source]` tuples through the split would make drift impossible regardless of future formatting changes.
---

PR #817, opened 2026-07-18
fix(llm): preserve database violation lifecycle

### LINKED ISSUE #791: `analyze`: no resume / partial re-run — a rate-limit near the end forces a full ~2.8M-token re-run

yahiaelmasry:
## Summary

- add a database-specific lifecycle schema and provider method so lifecycle results retain database and table targets
- strictly rebind current `db-*` prompt aliases while discarding unknown or wrong-namespace targets
- route prior database LLM findings through one lifecycle owner with fail-closed partition validation
- conservatively carry prior findings across skipped, malformed, or failed lifecycle work while keeping schema failures visible in progress

## Why

This is a correctness prerequisite for safe checkpointing and replay in #791. Database lifecycle results previously used the generic aggregate schema, which stripped `targetDatabaseId` and `targetTable`. The pipeline also ran the normal database path even when prior database LLM findings existed, so an old finding could be carried forward while an equivalent new finding was added.

The change gives database lifecycle work an explicit contract before those results can be journaled or reused. It does **not** implement durable run journals, checkpoint reuse, token preservation, Resume, usage reconciliation, latest-attempted-run pointers, or final promotion. It does not resolve or close #791.

PR #815 remains the separate session-limit damage-containment slice. If the two changes are merged in an order that causes conflicts in the provider or pipeline, both the terminal session-limit propagation from #815 and the database lifecycle behavior here must be preserved.

## Validation

- `pnpm exec vitest run tests/core/database-violation-lifecycle.test.ts tests/core/violation-lifecycle.test.ts tests/core/cli-provider.test.ts` — 31/31 passed
- `pnpm --filter @truecourse/core build` — passed
- fresh independent final audit — no actionable P0–P2 findings; its expanded focused matrix passed 37/37

Refs #791

### DIFF
--- a/packages/core/src/services/llm/cli-provider.ts
+++ b/packages/core/src/services/llm/cli-provider.ts
@@ -26,6 +26,7 @@ import {
 } from './prompts.js';
 import {
   ServiceViolationOutputSchema,
+  DatabaseLifecycleViolationOutputSchema,
   DatabaseViolationOutputSchema,
   ModuleViolationOutputSchema,
   DiffViolationOutputSchema,
@@ -53,6 +54,7 @@ import type {
   FlowEnrichmentResult,
   DiffViolationItem,
   DiffViolationsResult,
+  DatabaseViolationsLifecycleResult,
   ServiceDescription,
 } from './provider.js';
 
@@ -443,6 +445,40 @@ export abstract class BaseCLIProvider implements LLMProvider {
     };
   }
 
+  async generateDatabaseViolationsWithLifecycle(
+    context: DatabaseViolationContext,
+    opts?: { onStart?: () => void },
+  ): Promise<DatabaseViolationsLifecycleResult> {
+    const { vars, idMap } = buildDatabaseTemplateVars(context);
+    const prompt = getPrompt('violations-database-lifecycle', vars);
+
+    log.info('[CLI] Lifecycle database call starting...');
+    const t0 = Date.now();
+    const { data: object, usage: cliUsage } = await this.spawnAndParse(
+      prompt,
+      DatabaseLifecycleViolationOutputSchema,
+      {
+        extraArgs: ['--tools', ''],
+        label: 'database-lifecycle',
+        onStart: opts?.onStart,
+      },
+    );
+    const dur = Date.now() - t0;
+    log.info(`[CLI] Lifecycle database call done in ${dur}ms — resolved: ${object.resolvedViolationIds.length}, new: ${object.newViolations.length}`);
+    this.collectUsage('database', cliUsage, dur);
+
+    return {
+      resolvedViolationIds: resolveIds(object.resolvedViolationIds, idMap),
+      unchangedViolationIds: resolveIds(object.unchangedViolationIds, idMap),
+      newViolations: object.newViolations.map((violation) => ({
+        ...violation,
+        targetDatabaseId: violation.targetDatabaseId?.startsWith('db-')
+          ? idMap.get(violation.targetDatabaseId) ?? null
+          : null,
+      })),
+    };
+  }
+
   async generateModuleViolations(
     context: ModuleViolationContext,
     opts?: { onStart?: () => void },
@@ -543,6 +579,7 @@ export abstract class BaseCLIProvider implements LLMProvider {
   ): Promise<AllViolationsLifecycleResult> {
     const onCallStart = contexts.onCallStart;
     const allResolved: string[] = [];
+    const allUnchanged: string[] = [];
     const allNew: DiffViolationItem[] = [];
     let serviceDescriptions: ServiceDescription[] = [];
 
@@ -578,20 +615,9 @@ export abstract class BaseCLIProvider implements LLMProvider {
     if (contexts.database) {
       const ctx = contexts.database;
       if (ctx.existingViolations && ctx.existingViolations.length > 0) {
-        promises.push(['database', (async () => {
-          const { vars, idMap } = buildDatabaseTemplateVars(ctx);
-          idMaps.database = idMap;
-          const prompt = getPrompt('violations-database-lifecycle', vars);
-          log.info('[CLI] Lifecycle database call starting...');
-          const t0 = Date.now();
-          const { data: object, usage: cliUsage } = await this.spawnAndParse(prompt, DiffViolationOutputSchema, {
-            extraArgs: ['--tools', ''], label: 'database-lifecycle', onStart: () => onCallStart?.('database'),
-          });
-          const dur = Date.now() - t0;
-          log.info(`[CLI] Lifecycle database call done in ${dur}ms — resolved: ${object.resolvedViolationIds.length}, new: ${object.newViolations.length}`);
-          this.collectUsage('database', cliUsage, dur);
-          return object;
-        })()]);
+        promises.push(['database', this.generateDatabaseViolationsWithLifecycle(ctx, {
+          onStart: () => onCallStart?.('database'),
+        })]);
       } else {
         promises.push(['database-normal', this.generateDatabaseViolations(ctx, {
           onStart: () => onCallStart?.('database'),
@@ -620,6 +646,7 @@ export abstract class BaseCLIProvider implements LLMProvider {
           this.collectUsage('module', cliUsage, dur);
           return {
             resolvedViolationIds: resolveIds(object.resolvedViolationIds, idMap),
+            unchangedViolationIds: resolveIds(object.unchangedViolationIds, idMap),
             newViolations: object.newViolations.map((i) => {
               const realModuleId = resolveId(i.targetModuleId, idMap);
               return {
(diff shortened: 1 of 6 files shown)


DoilLee2222:
## Review & independent verification ✅

I checked out the PR head (`f4b225f`) on top of `main` (`ed625a5`), read the full changed sources (not just the diff), rebuilt `@truecourse/core`, ran the cited tests, **and wrote my own additional back-tests** to probe cases the shipped suite doesn't cover. Summary: the two stated bugs are genuinely fixed at the root, and everything that loads passes.

### What was verified

| Step | Command / action | Result |
|------|------------------|--------|
| Build (typecheck) | `pnpm --filter @truecourse/core... build` (9 pkgs, `tsc`) | ✅ passed — new schema/interface/pipeline types all check |
| Cited tests | `vitest run database-violation-lifecycle + violation-lifecycle + cli-provider` | ✅ **31/31** |
| Full core suite | `vitest run tests/core/` | ✅ **292/292** that load (see note on 2 unrelated suites) |
| CI | GitHub `test` job on this PR | ✅ success |

### Issue-fix confirmation (traced through the code)

**Bug ① — DB/table targets stripped in lifecycle mode.** Previously the lifecycle path parsed against `DiffViolationOutputSchema`, which has no `targetDatabaseId`/`targetTable`. This PR adds `DatabaseLifecycleViolationOutputSchema` + `generateDatabaseViolationsWithLifecycle`, so both survive. I confirmed the full round-trip closes: `targetDatabaseId` is preserved → persisted as `targetDatabaseName` (`analyze-persist.ts:312`) → remapped back to the new runtime id next run via `databaseNameToId` (`violation-lifecycle.service.ts:235`). Before this PR that chain was broken at step 1 (id nulled), so the name was never stored and cross-run remap could never work.

**Bug ② — double processing.** Database is now routed through exactly one owner: the arch path is called with `databases: undefined` / `existingDatabaseViolations: undefined` (`violation-pipeline.service.ts:1113,1126`) and `previousAggregateLlmViolations` explicitly **excludes** `type === 'database'` (`:1065`). So a prior DB finding can no longer be carried forward by arch *and* re-added by the DB path. This is structurally airtight, not just guarded by a flag.

**Fail-closed partition + alias rebinding.** `hasExactDatabaseLifecyclePartition` rejects any output where priors are omitted / duplicated / foreign, or where a "new" finding recreates an existing title; on rejection the batch throws and priors are carried forward unchanged. Alias rebinding correctly accepts only `db-*` ids (`db-99` → `null`, and a wrong-namespace `prev-0` → `null`). Verified against the `targets` / `overlap` / `outsider` / `*-duplicate` cases.

### Independent back-testing (gap I found)

Every shipped test uses **exactly one** prior violation — a `Set`-size or off-by-one partition bug would hide until 2+ priors. I wrote a throwaway suite with **two databases + two priors** and ran it green:

- **valid split** — 1 prior `resolved` + 1 `unchanged`, each remapped to its *own* db runtime id ✅
- **omission** — dropping one prior fails closed and carries **both** forward ✅
- **`db-1` rebind** — second-database alias resolves to the second runtime id (not hardcoded to `db-0`) ✅
- **title-recreate among many** — recreating one prior's title among several fails closed ✅

(Scratch test removed afterward; working tree left pristine.)

### Code-review notes (non-blocking)

1. **Dormant branch.** The database branch inside `generateAllViolationsWithLifecycle` (cli-provider) is currently unreachable from production — its only caller (`generateViolationsWithLifecycle`) always passes `databases: undefined`, so `dbContext` is never built. The PR keeps it type-consistent for the future, which is fine, but it's untested surface that could bit-rot; a short comment noting it's reserved for a future caller would help the next reader.
2. **Intentional divergence worth a comment.** The DB lifecycle treats a recreated-title finding as *invalid* (fail-closed), whereas `computeViolationLifecycle`'s generic path treats a same-title new finding as a *replacement* (drops the prior, keeps the new). Both are defensible; a one-line note on why DB is stricter would prevent a future "inconsistency" bug report.
3. **Progress count semantics.** The `"N violations"` detail counts `new + unchanged` and excludes `resolved`. That's correct (resolved == no longer active), just flagging it's deliberate.

### Note on the 2 failing suites (not this PR)

`tests/core/guard-read.test.ts` and `guard-decisions-overlay.test.ts` fail to **resolve** `@truecourse/ee-db` (an enterprise package not built in a plain checkout). I reproduced the identical failure on the base commit `ed625a5`, and neither test imports anything this PR touches — so these are pre-existing environmental failures, unrelated to #817.

### Verdict

Correct, well-scoped, and honest about what it does/doesn't do. No blocking issues found. The notes above are optional polish. 👍

---

PR #818, opened 2026-07-18
fix(llm): certify code result ownership

### LINKED ISSUE #791: `analyze`: no resume / partial re-run — a rate-limit near the end forces a full ~2.8M-token re-run

yahiaelmasry:
## Summary

Certify each LLM code-check result against the exact work unit that produced it before the result can enter violation lifecycle reconciliation.

- carry an explicit rule/source/range ledger through metadata, targeted, full-file, and oversized split batches;
- reject results that reference another batch's rule, file, range, or lifecycle IDs;
- require exact, duplicate-free lifecycle ID partitions and reject prior/new or duplicate-new collisions;
- preserve priors from failed or unclassified work units instead of inferring that they resolved;
- persist unchanged-only lifecycle results and count them in progress;
- keep metadata findings honestly file-level without copying whole source files into persisted snippets.

## Why

Safe checkpoint replay requires more than a stable work-unit ID. A saved result must also be provably owned by that work unit's exact inputs. Previously, concurrent code responses were flattened and validated only against the repository-wide file set, so a result could escape its originating batch and later be attached to the wrong rule, source, range, or prior lifecycle record.

This is a correctness prerequisite for the durable journal and resumability work planned in #791.

## Failure behavior

If a code batch fails schema or ownership certification, successful sibling calls may still report their findings, but the failed batch's prior findings are preserved as unchanged and the domain is shown as errored. Findings that no current certified work unit classified are also preserved. The pipeline does not treat absence of a certified response as evidence that a prior finding resolved.

## Tests

- 52 focused provider, ownership, router, batch-lifecycle, concurrency, and lifecycle tests
- `pnpm lint` — 14/14 tasks
- `pnpm build` — 19/19 tasks
- `git diff --check`
- fresh independent release audit: no P0–P2 findings

## Scope and follow-ups

Refs #791.

This PR does **not** resolve or close #791. It does not add stable work-unit IDs/fingerprints, a durable latest-attempted-run journal, token-preserving checkpoints, usage reconciliation, reset-aware resume, latest-run APIs/UI, or atomic completed-baseline promotion. Those remain required follow-up contributions in the accepted issue plan.

The current damage-containment tradeoff also remains: until the journal/checkpoint slices land, successful calls from an interrupted run are not reusable and a rerun can spend those tokens again.

## Interaction with PR #816

PR #816 and this PR both extend context-router provenance. If one lands first, the later rebase must preserve both contracts: #816's structured metadata/targeted/full-file source selections and this PR's exact rule/source/range result certification. The ownership ranges should be derived from the structured source selections; neither contract should replace the other.

### DIFF
--- a/packages/core/src/services/llm/cli-provider.ts
+++ b/packages/core/src/services/llm/cli-provider.ts
@@ -78,6 +78,94 @@ interface CLIUsage {
   costUsd?: string;
 }
 
+function certifyCodeResultOwnership(
+  context: CodeViolationContext,
+  violations: CodeViolationRaw[],
+): void {
+  const allowedRules = new Set(context.llmRules.map((rule) => rule.key));
+  const allowedSources = new Map<string, CodeViolationContext['sourceScopes'][number]['ranges']>();
+  for (const scope of context.sourceScopes) {
+    const ranges = allowedSources.get(scope.path) ?? [];
+    ranges.push(...scope.ranges);
+    allowedSources.set(scope.path, ranges);
+  }
+  const seenFindings = new Set<string>();
+
+  for (const violation of violations) {
+    if (!allowedRules.has(violation.ruleKey)) {
+      throw new Error(
+        `Code result rule "${violation.ruleKey}" is not owned by the originating batch`,
+      );
+    }
+
+    const ranges = allowedSources.get(violation.filePath);
+    if (!ranges) {
+      throw new Error(
+        `Code result source "${violation.filePath}" is not owned by the originating batch`,
+      );
+    }
+
+    const rangeIsOwned = violation.lineEnd >= violation.lineStart && ranges.some((range) =>
+      context.tier === 'metadata'
+        ? violation.lineStart === range.lineStart && violation.lineEnd === range.lineEnd
+        : violation.lineStart >= range.lineStart && violation.lineEnd <= range.lineEnd,
+    );
+    if (!rangeIsOwned) {
+      throw new Error(
+        `Code result range ${violation.lineStart}-${violation.lineEnd} for "${violation.filePath}" is not owned by the originating batch`,
+      );
+    }
+
+    const findingKey = `${violation.ruleKey}\0${violation.filePath}\0${violation.lineStart}\0${violation.lineEnd}\0${violation.title}`;
+    if (seenFindings.has(findingKey)) {
+      throw new Error('Code result cannot contain the same new finding more than once');
+    }
+    seenFindings.add(findingKey);
+  }
+}
+
+function certifyNoPriorCodeCollision(
+  context: CodeViolationContext,
+  newViolations: CodeViolationRaw[],
+): void {
+  const priorKeys = new Set((context.existingViolations ?? []).map((violation) =>
+    context.tier === 'metadata'
+      ? `${violation.ruleKey}\0${violation.filePath}\0${violation.title}`
+      : `${violation.ruleKey}\0${violation.filePath}\0${violation.lineStart}\0${violation.lineEnd}\0${violation.title}`,
+  ));
+
+  for (const violation of newViolations) {
+    const key = context.tier === 'metadata'
+      ? `${violation.ruleKey}\0${violation.filePath}\0${violation.title}`
+      : `${violation.ruleKey}\0${violation.filePath}\0${violation.lineStart}\0${violation.lineEnd}\0${violation.title}`;
+    if (priorKeys.has(key)) {
+      throw new Error(
+        'Code lifecycle result cannot classify a previous finding and reintroduce the same finding as new',
+      );
+    }
+  }
+}
+
+function certifyCodeLifecyclePartition(
+  resolvedViolationIds: string[],
+  unchangedViolationIds: string[],
+  idMap: PromptIdMap,
+): void {
+  const expectedIds = new Set(idMap.keys());
+  const classifiedIds = [...resolvedViolationIds, ...unchangedViolationIds];
+  const classifiedSet = new Set(classifiedIds);
+  const isExactPartition =
+    classifiedIds.length === expectedIds.size &&
+    classifiedSet.size === expectedIds.size &&
+    classifiedIds.every((id) => expectedIds.has(id));
+
+  if (!isExactPartition) {
+    throw new Error(
+      'Code lifecycle result must be an exact partition of the originating batch previous IDs',
+    );
+  }
+}
+
 export abstract class BaseCLIProvider implements LLMProvider {
   abstract get binaryName(): string;
   abstract get baseArgs(): string[];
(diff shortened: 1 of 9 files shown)


DoilLee2222:
## Verification + code review

I checked this PR out at its head commit (`05209ee`) in an isolated worktree, built it, ran the tests, traced the core logic by hand, and ran an additional set of scenario tests I wrote independently.

**Verdict: ✅ The change does what it claims.** Build is green, the full core suite passes, and every ownership/lifecycle invariant I checked holds. No correctness bugs found. Two minor, non-blocking observations are at the end.

> Note: this is a prerequisite slice of #791 and does not close it — verification is against the PR's stated scope, not full resumability.

### What I ran

| Check | Result |
|---|---|
| `pnpm build` (all packages, tsc) | ✅ pass |
| Targeted suite — ownership, batch-lifecycle, router, provider, concurrency, id-mapping, lifecycle | ✅ 66/66 |
| Full `tests/core` suite | ✅ 345/345 |
| Independent scenario tests I added | ✅ 9/9 |

Building clean is itself meaningful here: `sourceScopes` became a **required** field on `CodeViolationContext`, and I confirmed every construction site (only the pipeline) sets it — no caller left behind.

### Logic traced by hand

- **Single certification path.** `BaseCLIProvider` is the only implementation of `generateCodeViolations`; all success / lifecycle / oversized-split responses pass through the three gates (rule+source+range ownership, exact prior-ID partition, prior/new collision). No bypass. `generateAllCodeViolations` (the older flatten-and-drop path) has no remaining callers.
- **Batch-split provenance alignment.** `content.split(/(?=^=== )/m)` yields one section per file with no leading empty element (verified empirically), and `sourceScopes[si]` is pushed 1:1 with `parts[si]`, so scopes stay attached to the correct split batch.
- **Failure preservation (the core guarantee).** I simulated the real pipeline composition — two concurrent batches where one fails, one succeeds and resolves a prior, plus a prior in an unscanned file. `aggregateCodeBatchOutcomes` → `reconcileCodePriorClassifications` preserved the failed-batch prior and the unscanned prior as `unchanged`, honored the sibling resolve, and accounted for every prior exactly once. Absence of a certified response is never treated as "resolved," exactly as claimed.
- **No double classification.** `reconcile` drops any id from `resolved` that also appears in `unchanged` (unchanged wins), and downstream `handledIds` covers all prior LLM ids, so a prior can never surface as both resolved and unchanged.
- **Metadata honesty.** Fabricated narrow ranges are rejected; only exact file-level bounds are accepted (`===`-exact match in metadata tier); `buildLlmCodeSnippet` persists an empty snippet for metadata-tier findings.

### Code-review notes (non-blocking)

1. **Shrunk-file priors.** If a file shrinks so a prior's stored range now exceeds the new line count, that prior is excluded from its batch's lifecycle input and carried forward as `unchanged` by the reconcile fallback. In the rare case the same issue is also re-reported fresh, you could momentarily see a stale `unchanged` card next to the new finding. This is over-preservation, not data loss, and matches the PR's stated damage-containment tradeoff — flagging only so it's a conscious choice.

2. **Dead positional-resolve branch.** After the reconcile pass, `handledIds` covers every prior LLM violation, so `llmPrevForMatching` passed to `computeFileViolationLifecycle` is always empty. That call now only inserts new findings; its positional resolve/unchanged matching is effectively unreachable for the LLM code path. Behavior is correct, but the empty-`previousViolations` call is a cleanup opportunity.

### Re: interaction with #816

Both this PR and #816 extend context-router provenance and touch the same `buildMetadata/Targeted/FullFileContent` seams. Whichever merges second should keep both #816's structured source selections and this PR's `sourceScopes` ownership ranges — deriving the latter from the former rather than replacing it.


---

PR #819, opened 2026-07-20
Fix #813: C# Roslyn host runs on the SDK's runtime (analyze-project MSBuildLocator failure)

### LINKED ISSUE #813: C# semantic analysis aborts: Roslyn host runs on .NET 8 but only a .NET 10 SDK is installed → MSBuildLocator finds no MSBuild
## Summary

The project-aware C# tier (`engine: 'roslyn-workspace'`, op `analyze-project`) fails **fatally** in the `cs-` scope environment because `MSBuildLocator.RegisterDefaults()` cannot find an MSBuild instance. This aborts the entire `truecourse analyze` run (the workspace-tier failure is rethrown as a fatal error in `violation-pipeline.service.ts`), so **no `LATEST.json` is produced** and every downstream step is blocked.

Discovered by the `cs-` fp-next-fix routine while analyzing `OrchardCMS/OrchardCore` @ `1d0ae14413f10e2b3c1a65927db463b30a6892fe`.

## Root cause

- The Roslyn host DLL (`dist/roslyn-host/csharp-roslyn-host.dll`) targets **net8.0** with `runtimeOptions.rollForward: "Major"` and `framework.version: 8.0.0`.
- `setup-csharp-env.sh` provisions a **.NET 10 SDK (10.0.302)** + a **.NET 8 runtime**, but **no .NET 8/9 SDK**.
- Because a net8 runtime *is* present, the host runs on .NET 8 (roll-forward Major only jumps majors when no net8 is available). `MSBuildLocator` in a net8 process cannot use the net10 SDK's MSBuild → `"No instances of MSBuild could be detected."`

## Evidence (this environment)

```
$ dotnet --list-sdks
10.0.302 [/root/.dotnet/sdk]           # only a net10 SDK
$ dotnet --list-runtimes
Microsoft.NETCore.App 8.0.29 ...       # net8 runtime present → host runs on net8
Microsoft.NETCore.App 10.0.10 ...
```

Host on default (net8) runtime — **fails**:
```
$ echo '{"op":"analyze-project","project":".../repro.csproj","rules":[...]}' \
    | dotnet exec dist/roslyn-host/csharp-roslyn-host.dll
{"ok":false,"error":"could not locate a .NET SDK to open the project via MSBuild:
 No instances of MSBuild could be detected. ..."}
```

Same request, host forced onto the net10 runtime — **succeeds**:
```
$ echo '{"op":"analyze-project",...}' \
    | dotnet exec --fx-version 10.0.10 dist/roslyn-host/csharp-roslyn-host.dll
{"ok":true,"violations":[ ... ]}
```

## Impact

- All `roslyn-workspace` rules are blocked; because the failure is fatal, the whole analyze aborts (no results at all).
- Blocks the entire `cs-` campaign pipeline (fp-discover / fp-next-fix / queue-empty re-measure) on any repo that reaches the C# semantic phase.

## Suggested fix (pick one)

1. **Run the host on the SDK's runtime.** Bump the host TFM to net10 (or set `rollForward: LatestMajor` / launch with `--fx-version` matching the installed SDK) so MSBuildLocator can use the net10 SDK. This keeps the host runtime in lock-step with whatever SDK `setup-csharp-env.sh` installs.
2. **Install a matching SDK.** Have `setup-csharp-env.sh` also install a .NET 8 (or 9) SDK so the net8 host finds a compatible MSBuild. (Heavier; option 1 is cleaner and avoids drift when target repos pin newer SDKs via `global.json`.)

Whichever is chosen, `setup-csharp-env.sh` and the host runtimeconfig should be kept in sync so the host's runtime and the installed SDK's MSBuild always match.

## Notes

- This is separate from, and independent of, the deterministic-phase hang tracked in the sibling issue — both must be resolved for a `cs-` whole-repo analyze to complete.
- Not an fp-fix; belongs in build/tooling, not an `fp-fix` batch.

DoilLee2222:
## Summary

Fixes #813. The project-aware C# tier (`engine: 'roslyn-workspace'`, op `analyze-project`) aborted **fatally** in the `cs-` environment: the host is a `net8.0` DLL and, with `RollForward=Major` preferring an installed net8 runtime, ran on net8 while only a net10 SDK was present, so `MSBuildLocator` found no compatible MSBuild (`"No instances of MSBuild could be detected."`). That failure is fatal in the workspace tier, so no `LATEST.json` was produced and the whole analyze aborted.

## Root cause

A single host process had **two conflicting runtime requirements**:

- The loose-text `analyze` op derived its reference set from the *running runtime's* `TRUSTED_PLATFORM_ASSEMBLIES` (the runtime's BCL). That is why the host was pinned to net8 — to keep the BCL deterministic for version-sensitive semantic rules (e.g. the `CallerArgumentExpression` overload of `Debug.Assert`).
- The `analyze-project` op needs the process runtime major to match the installed SDK's runtime so `MSBuildLocator` can register it.

net8 (for loose-text) and net10 (to match the SDK) cannot both be satisfied by one process — so `analyze-project` broke.

## Fix (decouple the two requirements at the root)

1. **`BuildReferences()` no longer uses the live runtime's TPA.** It uses a pinned net8 reference-assembly pack (`Basic.Reference.Assemblies.Net80`), so loose-text results are deterministic regardless of which runtime the host runs on. `ILogger` / `Newtonsoft.Json` references (previously implicit via TPA) are added explicitly from their pinned package assemblies.
2. **`RollForward=Major` → `LatestMajor`.** The host now runs on the SDK's own runtime, so `MSBuildLocator` always finds a usable MSBuild. This is safe now because the BCL no longer follows the process runtime — the regression that previously forced a net8 pin is gone.
3. **`setup-csharp-env.sh` no longer installs a standalone net8 runtime.** The host rolls forward onto the SDK's runtime, removing the exact net8-runtime + net10-SDK trap that triggered the failure.
4. **Docs / comments** updated to match (host README, fp-automation README, one test comment).

The host runtime and the installed SDK now stay in lock-step by construction, and target repos that pin newer SDKs via `global.json` are tracked automatically.

## Not changed (by design)

When the required SDK is *genuinely absent* (e.g. a net8-only box analyzing a net10-pinned repo), `analyze-project` still fails **loudly** with an actionable message (`"Restore the project (dotnet restore) and ensure its pinned .NET SDK is installed."`). That is the intended fail-hard behavior — a missing SDK is an operator/provisioning issue, not something the host can synthesize, and a silent half-analysis is worse than a clear error. This is a distinct concern from #813 (where the correct SDK *was* installed but unusable) and is intentionally left as-is.

## Verification

- Confirmed the exact API by unpacking the NuGet package: `Basic.Reference.Assemblies.Net80.References.All` (→ `ImmutableArray<PortableExecutableReference>`).
- CI (`test.yml` / `publish.yml`) builds the host and runs the Roslyn tests on a net8 SDK, which compile-checks the new `BuildReferences` code and exercises both analyze ops. CI does not reproduce the exact net10-SDK / net8-runtime mismatch (that is the production `cs-` environment); the structural fix covers it. A net10 CI matrix could be added later as a regression guard.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QJkekx28SQ4S6M3xUNn2xr)_

### DIFF
--- a/docs/fp-automation/README.md
+++ b/docs/fp-automation/README.md
@@ -298,8 +298,11 @@ support to produce findings; the `drift-fp` loop verifies C# via the contract-ve
 **C# account .NET prerequisites** (default account needs none):
 - The session must have the **.NET SDK — 10.x** recommended (or ≥ 9.0.2xx). It
   builds the Roslyn host during `build:dist` and, at analyze time, opens `.slnx`
-  solutions (common in modern C# repos) and modern targets. The host itself is
-  `net8.0` with roll-forward, so it *runs* on any ≥ 8 runtime.
+  solutions (common in modern C# repos) and modern targets. The host itself is a
+  `net8.0` DLL with `RollForward=LatestMajor`, so it *runs* on the newest runtime
+  present — deliberately the SDK's own runtime, so `analyze-project`'s
+  MSBuildLocator always finds a usable MSBuild. Loose-text analysis stays on a
+  pinned net8 reference set regardless of that runtime (issue #813).
 - Each C# target must be **`dotnet restore`d before analyze** (see the discover /
   next-fix prompts). The project-aware Roslyn tier **fails-hard** on an
   unrestored project or a missing/unsatisfiable SDK — it aborts the run rather
@@ -645,9 +648,9 @@ One-time, before the first run:
      .NET CDN — so a mid-session install returns 403 from the egress proxy):
      1. **Setup script.** Give the C# account its **own** environment whose
         setup script runs `docs/fp-automation/setup-csharp-env.sh` — it
-        installs the 10.x SDK plus a net8 runtime for the host, once per
-        container, and is idempotent. (The routine prompts also run it as a
-        fallback when `dotnet` is missing.)
+        installs the 10.x SDK (whose runtime the `net8.0` host rolls forward
+        onto), once per container, and is idempotent. (The routine prompts also
+        run it as a fallback when `dotnet` is missing.)
      2. **Network policy.** Widen that environment's egress allowlist to the
         .NET download + NuGet hosts — at least `dot.net`,
         `builds.dotnet.microsoft.com`, `dotnetcli.azureedge.net`,
--- a/docs/fp-automation/setup-csharp-env.sh
+++ b/docs/fp-automation/setup-csharp-env.sh
@@ -45,13 +45,15 @@ fi
 chmod +x /tmp/dotnet-install.sh
 
 # .NET 10 SDK — abp's global.json pins 10.0.100 (rollForward latestFeature); the
-# roslyn-workspace tier restores abp's real solution with this SDK.
+# roslyn-workspace tier restores abp's real solution with this SDK. Installing the
+# SDK also lays down its matching runtime, which is all the host needs: the host is
+# a net8.0 DLL with RollForward=LatestMajor, so it runs on this SDK's runtime, and
+# `analyze-project`'s MSBuildLocator finds the SDK's MSBuild (running on a separate
+# net8 runtime while only a net10 SDK is present cannot — issue #813). No standalone
+# net8 runtime is installed: the loose-text `analyze` op no longer depends on the
+# runtime BCL (it uses a pinned net8 reference pack), so the host is free to run on
+# the SDK's runtime.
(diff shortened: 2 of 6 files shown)


spiderhands (approved):
I've verified the fix. Nicely done!

One thing though - this should be merged after #808

PR #820, opened 2026-07-20
fix(analyzer): isolate deterministic scan in a killable worker so one file can't hang analyze (#814)

### LINKED ISSUE #814: analyze hangs indefinitely on a pathological file in the deterministic (tree-sitter) phase — one bad file freezes the whole run
## Summary

`truecourse analyze` on a large repo can hang **indefinitely** in the deterministic (tree-sitter) phase. A single source file drives one deterministic rule into effectively non-terminating work, and because the phase waits on all files, the entire run stalls — no `LATEST.json` is ever written.

Discovered by the `cs-` fp-next-fix routine analyzing `OrchardCMS/OrchardCore` @ `1d0ae14413f10e2b3c1a65927db463b30a6892fe`.

## Observed behavior

- Progress froze at exactly `4320/5551 files` and never advanced for **80+ CPU-minutes** (killed manually; it was not going to finish).
- The node process sat pinned at ~115–120% CPU the whole time with `wchan = 0` (busy in userspace, **not** blocked on I/O, a lock, or the network).
- Memory (~5 GB RSS, ~9 GB free) and disk were fine → not GC thrash, not OOM, not disk-full.

This signature — one file's *content*, not its size, causing super-linear/explosive work in a single rule — is the classic shape of **catastrophic regex backtracking** in a deterministic pattern.

## What was ruled out

- **File size alone is not it.** The largest minified bundles are skipped (`*.min.js` → 0 files) or complete fine individually and in a batch (e.g. `graphiql-orchard.js` ~1.1 MB and fontawesome `all.js` ~1.7 MB both scanned in seconds).
- **Not network / not a lock / not resources** (see above).

## Not yet isolated

The exact **file + rule** were not pinpointed this session (that needs a re-run with per-file tracing, and a second independent blocker — the net8/net10 MSBuild issue in the sibling issue — made further investigation lower priority). The frozen counter (`4320`) means the offending file was in the in-flight batch while its peers had already completed.

## Impact

- Any sufficiently large/varied repo can hang the whole analyze on one file → blocks fp-discover / fp-next-fix and any interactive analyze.

## Suggested fix (two layers)

1. **Root cause (primary):** identify the file + rule (re-run with per-file logging of which path each deterministic rule is scanning, or bisect the file set), then fix the offending pattern so it can't backtrack catastrophically (anchor / possessive / bounded quantifiers, or replace the regex with a linear scan). Aligns with the repo's "find and fix the root cause" rule.
2. **Safety net (defense in depth):** add a per-file time (or size) guard to the deterministic phase so a single pathological file is **skipped with a warning** instead of freezing the entire run. A bounded, logged skip is far better than an unbounded hang.

## Repro pointer

Analyze `OrchardCMS/OrchardCore` @ `1d0ae14413f10e2b3c1a65927db463b30a6892fe` with `analyze --no-llm --no-stash --no-skills`; it stalls at `4320/5551` in the deterministic phase. (Note: to reach a full run you'll also need the sibling net8/net10 MSBuild fix, but the hang occurs *before* the C# semantic phase.)

## Notes

- Independent of the net8/net10 MSBuild blocker (sibling issue); both must be fixed for a `cs-` whole-repo analyze to complete.
- Not an fp-fix; belongs in analyzer robustness, not an `fp-fix` batch.

DoilLee2222:
## Summary

Fixes #814. `truecourse analyze` could hang **indefinitely** in the deterministic (tree-sitter) phase: a single source file drove a rule's regex into catastrophic backtracking, and because the per-file loop ran **synchronously on the main thread**, the whole run froze — no `LATEST.json` was ever written, and even `Ctrl+C` had no effect (the main thread never got a chance to check the abort flag).

The root cause is structural, not one bad regex: nothing could interrupt a single file's synchronous work. You cannot interrupt a running V8 regex from the same thread, so a real fix has to move the work across a thread boundary. This PR does that, and adds a regression gate so a provably-exponential pattern can't be introduced later.

### Layer 1 — runtime guarantee (killable worker + per-file timeout)

The per-file tree-sitter scan now runs in a **worker thread** with a per-file wall-clock budget:

- A file that exceeds the budget is **terminated and skipped-with-a-warning** (the warning names the file), and the scan **resumes past it**, so the run always completes and writes its output.
- Because the main thread is now free, the watchdog fires reliably **even while the worker is pinned in native regex backtracking**, and `Ctrl+C` takes effect mid-scan.
- The budget defaults to **30s/file** and is overridable via `TRUECOURSE_DET_FILE_TIMEOUT_MS` (a universal resource bound, not a file-specific heuristic).
- Falls back to an in-thread scan when no worker entry can be resolved, so `analyze` never breaks.

This is input-agnostic: it bounds exponential backtracking, quadratic blowup, dynamically-built regexes, and any future non-regex CPU pathology alike. The skip warning also names the offending file, which is exactly the per-file tracing the issue said was still needed to pinpoint the culprit.

### Layer 2 — regression prevention (ReDoS gate)

A test enumerates the analyzer's regex **literals** (via the TypeScript compiler) and fails the build on any that are provably vulnerable to **exponential** backtracking (`redos-detector`). The current corpus has no such literal; the gate blocks future regressions. A cheap structural pre-filter keeps it fast (~6s). Documented limits: it can't see dynamic `new RegExp(...)` patterns and doesn't flag polynomial blowup — both are bounded at runtime by Layer 1.

## Changes

- **New module** `packages/analyzer/src/deterministic-scan/` (`types`, `run-scan`, `worker`, `controller`), exported from the analyzer index. `TypeQuery` / schema-index construction moves into the scan (they were loop-local).
- `violation-pipeline.service.ts` calls `runDeterministicScanIsolated` in place of the old synchronous loop, preserving the progress counter, abort behavior, and violation output.
- `scripts/build.ts` emits `dist/det-scan-worker.mjs` alongside the entry bundles (same externals + shared `dist/wasm` dir); the controller resolves the worker for both bundled and unbundled layouts.
- `README.md` documents the per-file budget and the env var.

## Testing

- **New** `tests/analyzer/deterministic-scan.test.ts`: clean run, stall → skip → resume, multiple stalls, mid-scan abort, and the in-thread fallback. The stall fixture drives a **real catastrophic regex** (`(a+)+$`), proving `worker.terminate()` kills a thread stuck in the native regex engine — the actual failure mode of #814.
- **New** `tests/analyzer/redos-regex-gate.test.ts`: the ReDoS gate.
- Verified the real built worker end-to-end (WASM init + TypeQuery build inside the worker) produces correct violations with `skipped=0`.
- Full suites green: analyzer + server (4615 passed), CLI e2e `analyze` for JS/TS + Python, and shared. The only failing test is the pre-existing `analyze-csharp` e2e, which requires the .NET SDK (`spawnSync dotnet ENOENT`) and is unrelated to this change.

## Note for reviewers

The **bundled** worker (`dist/det-scan-worker.mjs`) was confirmed to build via esbuild, but its runtime module resolution could not be fully exercised in this environment (that needs a full `pnpm build:dist`, i.e. `npm install` inside `dist/`). It uses the exact same externals, banner, output location, and WASM-dir convention as the already-shipping `cli.mjs`, and the identical *unbundled* worker was run end-to-end — so it is correct by construction, but a `pnpm build:dist` + analyze on a large repo in CI is worth a final check before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ TrueCourse ships with **1,200+ deterministic rules** and **100 LLM rules** acros
 
 **Deterministic rules** run via tree-sitter AST visitors — fast, zero-cost, no API calls. **LLM rules** send source code to the configured LLM for semantic analysis — deeper but requires an LLM provider.
 
+The deterministic scan runs in a worker thread with a per-file time budget, so a single pathological file (e.g. one that drives a rule's regex into catastrophic backtracking) is skipped-with-a-warning instead of freezing the whole run — the analysis always completes and writes its output. The budget defaults to 30s per file and is overridable with `TRUECOURSE_DET_FILE_TIMEOUT_MS` (milliseconds) for repos with unusually large legitimate sources.
+
 ## Commands
 
 ```bash
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "react-router-dom": "^7.1.0",
+    "redos-detector": "^6.1.4",
     "sonner": "^2.0.7",
     "supertest": "^7.2.2",
     "tsx": "^4.19.0",
--- a/packages/analyzer/src/deterministic-scan/controller.ts
+++ b/packages/analyzer/src/deterministic-scan/controller.ts
@@ -22,18 +22,30 @@ export interface DeterministicScanResult {
 export interface ScanControllerOptions {
   /** Per-file wall-clock budget. A file exceeding it is terminated and skipped. */
   fileTimeoutMs: number
+  /**
+   * Budget for the worker's one-time setup (WASM parser init + whole-project
+   * TypeScript program build) — the phase before any file is processed. Bounds
+   * a hang there so it can't freeze the run the way per-file work can't.
+   * Generous by design; defaults to {@link DEFAULT_SETUP_TIMEOUT_MS}.
+   */
+  setupTimeoutMs?: number
   /** Cooperative cancellation (e.g. Ctrl+C). */
   signal?: AbortSignal
   /** Progress reporter — `processed` counts files finished OR skipped, out of `total`. */
   onProgress?(processed: number, total: number): void
   /** Called once per skipped file so the caller can warn. */
   onSkip?(filePath: string, reason: string): void
+  /** Called when no worker could be resolved and the scan runs in-thread (no hang protection). */
+  onFallback?(): void
   /**
    * Override the worker entry path. Normally resolved automatically for the
    * current packaging layout; provided for tests and advanced embedders.
    */
   workerPath?: string
 }
+
+/** Default worker-setup budget — generous, since a large repo's type program build takes minutes. */
+export const DEFAULT_SETUP_TIMEOUT_MS = 600_000
 
 /** Input for the controller — `startIndex` is managed internally across resume passes. */
 export type IsolatedScanInput = Omit<DeterministicScanInput, 'startIndex'>
@@ -78,21 +90,40 @@ function runOneWorkerPass(
     let inFlightIndex = input.startIndex
     let watchdog: ReturnType<typeof setTimeout> | undefined
 
+    const clearWatchdog = () => {
+      if (watchdog) {
+        clearTimeout(watchdog)
+        watchdog = undefined
+      }
+    }
+
     const finish = (outcome: PassOutcome) => {
       if (settled) return
       settled = true
-      if (watchdog) clearTimeout(watchdog)
+      clearWatchdog()
       opts.signal?.removeEventListener('abort', onAbort)
       void worker.terminate()
       resolve(outcome)
     }
 
-    // The watchdog only runs while a file is in flight. It is (re)armed on every
-    // message, so a file that stops producing messages for `fileTimeoutMs` is
-    // deemed stalled. The main thread is otherwise idle, so this timer fires
-    // reliably even while the worker is pinned in synchronous backtracking.
-    const armWatchdog = () => {
-      if (watchdog) clearTimeout(watchdog)
+    // Every phase of the worker's life is covered by a timer, so no hang can go
+    // unbounded. The main thread is otherwise idle, so these timers fire
+    // reliably even while the worker is pinned in synchronous work (regex
+    // backtracking, type-checking):
+    //  - setup (WASM parser init + whole-project TS program build) is bounded by
+    //    `setupTimeoutMs`, armed from spawn until the `setup-done` message.
+    //  - each file is bounded by `fileTimeoutMs`, (re)armed on every message.
+    const setupTimeoutMs = opts.setupTimeoutMs ?? DEFAULT_SETUP_TIMEOUT_MS
+    const armSetupWatchdog = () => {
+      clearWatchdog()
+      watchdog = setTimeout(
+        () => finish({ kind: 'error', message: `worker setup (parser init + type program build) exceeded ${setupTimeoutMs}ms` }),
+        setupTimeoutMs,
+      )
+      watchdog.unref?.()
+    }
+    const armFileWatchdog = () => {
+      clearWatchdog()
       watchdog = setTimeout(() => finish({ kind: 'stall', stalledIndex: inFlightIndex }), opts.fileTimeoutMs)
       watchdog.unref?.()
     }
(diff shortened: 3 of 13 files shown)


spiderhands (approved):
Really solid work — this is the right architecture for the problem, and you clearly understood *why* the bug was unfixable from the same thread: you can't interrupt synchronous JS (or a running regex) except by killing the thread it runs on. Streaming results per file so nothing is lost when a worker is killed, resuming past the bad file, and proving the kill works against a real `(a+)+$` regex in the test are all exactly right. Four things to address or think about before merge:

**1. The worker's startup has no timeout — so a hang there can still freeze the run forever.**

Your watchdog timer only starts counting when the first `file-start` message arrives. But before the worker sends any message at all, it does two things: loads the tree-sitter WASM parsers, and calls `createTypeQueryService`. That second call builds a full TypeScript program for the whole repo up front (look at `ts-compiler.ts:411` — `ts.createProgram` runs right inside it, not lazily on first use). On a big repo this is the slowest part of the entire analysis — minutes, not seconds. If it ever gets stuck (weird tsconfig, symlinked directory cycles), no message is ever sent, no watchdog is ever armed, and we're back to the exact bug this PR fixes: an infinite hang.

Fix idea: have the worker send a `setup-done` message, and give the setup phase its own (generous — say 5–10 min) timeout in the controller. Then every phase of the worker's life is covered by *some* timer.

**2. Each skipped file re-pays the whole startup cost.**

When a file stalls, you terminate the worker and start a fresh one — which re-runs that whole TypeScript program build from scratch. One bad file = fine. Five bad files on a repo where the program build takes 3 minutes = ~15 extra minutes of wall clock. The run still finishes (which is the important thing!), so this doesn't block merge — but it's worth a comment in the code acknowledging the trade-off, so the next person doesn't think it was an accident.

**3. The fix contains the damage but doesn't cure the sick files — and those files now silently lose analysis coverage.**

Here's something subtle: your ReDoS gate itself proves there's *no* exponential-backtracking regex literal in the codebase. So what actually made #814 hang was almost certainly not catastrophic backtracking — it was quadratic visitors. For example, `floating-promise.ts:20` calls `sourceCode.split('\n')` — which re-splits the *entire file* — once for *every expression statement in the file*. On a 100k-line generated file that's O(statements × file-size): hours of work, looks exactly like a hang. `makeViolation` in `rules/types.ts:46` has the same problem (full re-split per violation).

With your PR, such a file now burns its 30s budget and gets skipped — so we stop hanging, but we also stop analyzing that file at all. The actual cure is tiny: split the file into lines **once** per file and pass the array down. Then those files scan in milliseconds and never hit the timeout. I'd suggest this as your follow-up PR — it's a great complement to this one (this PR = the safety net, that PR = fixing the reason we fall into it).

**4. Two smaller things:**

- If `resolveWorkerPath()` returns null, we silently fall back to the old in-thread scan — which means silently losing the whole hang protection. Add a `log.warn` there so we can tell from a user's log which mode they were running in.
- Skipped files only go to the log. Consider putting the skip list into the scan result / snapshot too, so the dashboard could eventually show "2 files skipped" instead of it being invisible.

And one process note: your own PR description flags that the bundled `det-scan-worker.mjs` was never executed end-to-end. Please do run `pnpm build:dist` + an analyze on a real repo before merging — if the worker file doesn't resolve in the bundled layout, every production user silently drops to the unprotected fallback, and nothing would tell us.

Again — genuinely good PR. The core design (killable worker, streamed results, resume, real-regex test) is exactly how this should be done. 👏

DoilLee2222:
Thanks for the thorough review — great catches. Addressed points 1, 2, and 4 in `42a4a24`:

**1. Setup-phase timeout.** You're right that the whole setup window (WASM init + `ts.createProgram`) was uncovered — a hang there would reproduce the exact bug. The worker now posts a `setup-done` message; the controller arms a generous **setup watchdog** (default 10m, `TRUECOURSE_DET_SETUP_TIMEOUT_MS`) from spawn until that message, then switches to the per-file budget. So every phase of the worker's life is now bounded by *some* timer. Added a test that pins the thread *during* setup and asserts the setup timer (not the per-file one) catches it.

**2. Resume re-pays setup cost.** Documented as an accepted trade-off with a comment at the resume site — and noted why keeping a worker alive across a stall isn't possible (the isolate holding the type program is exactly what we must terminate to unstick it).

**4a.** The silent in-thread fallback now fires an `onFallback` callback → the pipeline emits a clear `log.warn`, so a user's log shows when they ran without hang protection.

**4b.** `ViolationPipelineResult` now carries `skippedDeterministicFiles` so the skip list is available to callers/the dashboard, not just the log.

**Point 3 (the real cure — quadratic visitors like `floating-promise.ts` / `makeViolation` re-splitting the whole file per statement)** — this is a genuinely sharp observation, and I agree it's the actual reason #814 hung (my ReDoS gate already proves there's no exponential-backtracking literal). I'm taking it as the **follow-up PR** you suggested: this PR is the safety net; that one splits each file into lines once and removes the reason we hit the timeout at all.

Still outstanding before merge: your process note about running `pnpm build:dist` + an analyze on a real repo to confirm the bundled `det-scan-worker.mjs` resolves in the bundled layout. I couldn't fully exercise that here (no `dist/node_modules`); it uses the identical externals/banner/location as `cli.mjs`, but it's worth a real check.

---

DoilLee2222:
Thanks for the thorough review, @spiderhands. I'm handling item **#3** (the quadratic per-node re-splitting that makes files fall into the timeout, e.g. `floating-promise.ts:20` and `makeViolation`) in a separate follow-up PR: #823.

That PR keeps this one as the safety net and removes the underlying cause — it splits each file into lines once and reuses it (via a memoized `splitLines` helper), so those files scan in milliseconds instead of burning their timeout budget and getting skipped. This PR stays focused on the killable-worker safety net.

---

Issue #832, opened 2026-07-23
[cs-fp-campaign-stuck] OrchardCMS/OrchardCore


DoilLee2222:
The `OrchardCMS/OrchardCore` C# campaign cannot self-progress through the automated fp-next-fix routine. This is the human-signal issue; details below.

The fp-next-fix routine ran this session (SCOPE=`cs-`, TECH_STACKS=`csharp`) and found **both pickable queues empty** — every open `cs-fp-fix` issue for this campaign already carries `cs-fp-skipped` (see checklist). Per the queue-empty path, it re-measured the campaign, but the re-measure is blocked (Blocker 1), and the remaining issues can't be fixed by automation (Blocker 2).

## Where the campaign stands

- **tp_rate: unmeasurable this session** (see Blocker 1). Close threshold is **0.90**.
- Target: `OrchardCMS/OrchardCore` @ `1d0ae14413f10e2b3c1a65927db463b30a6892fe` (the ref all open `cs-fp-fix` issues cite; `campaigns.yaml` still has `status: discovering` with a null baseline).
- Analyze this session ran on a freshly-built dist (`pnpm build:dist` including the just-merged #831 fixes) via `node dist/cli.mjs analyze --no-llm --no-stash --no-skills`. It completed **every** rule category — **95,320 violations across 5,551 files** (Code quality 55,152 · Style 21,345 · C# semantic 12,397 · Bugs 5,019 · Architecture 1,068 · Security 425 · Performance 327 · Reliability 227 · Database 3) — then crashed while materializing the result.

## Blocker 1 — the whole-repo analyze crashes on this target, so no re-measure is possible

After computing all 95,320 violations, the run dies at the "Saving results" step with:

```
x  Invalid string length
```

i.e. V8's `RangeError: Invalid string length` — the serialized OrchardCore result JSON exceeds V8's maximum string length, so **no `LATEST.json` / snapshot is written** and per-rule counts cannot be produced. This is a target-scale limit in the store's JSON serialization, independent of any rule change (a small C# project serializes fine on the same dist). It is the same crash documented in **#829** and reproduced in **#831** (whose FP-count-delta section is `unavailable` for exactly this reason). It reproduced **again this session**.

Consequences for automation:
[...]

PR #831, opened 2026-07-22
fix(fp): resolve 3 FPs from OrchardCMS/OrchardCore

DoilLee2222:
## Human review required

These fixes ship in a **human-review** batch: the entire pickable `cs-fp-fix` queue carried `cs-fp-human-review-needed`, so there was no normal-phase queue this session. `fp-next-fix-review` will review but **not** auto-merge — a human must review and merge.

Non-rule modules touched (over the in-bounds boundary):
- **Roslyn semantic host** (`tools/csharp-roslyn-host/Rules/CodeQuality/StaticHolderTypeNotSealed.cs`) — for #782.

#777 and #781 are **in-bounds** (rule visitors + fixtures only) but ship in this human-review batch because there was no normal queue to route them to.

Closes #777
Closes #781
Closes #782

## Fixes

| rule_key | issue | positive-fixture | negative-fixture |
|---|---|---|---|
| architecture/deterministic/uri-property-as-string | #777 | `sample-csharp-project-positive/services/ArchitectureSamples/BookmarkFieldSafe.cs`, `BookmarkIndexSafe.cs` | `sample-csharp-project-negative/services/UserService/Violations/Architecture/FeedSubscription.cs` |
| code-quality/deterministic/no-extraneous-class | #781 | `sample-csharp-project-positive/services/CodeQualitySamples1/ProtectedHelperBaseSafe.cs` | `sample-csharp-project-negative/services/UserService/Violations/CodeQuality/GeometryHelpers.cs` |
| code-quality/deterministic/static-holder-type-not-sealed | #782 | `sample-csharp-project-positive/services/CodeQualitySamples1/ProtectedHelperBaseSafe.cs` | `sample-csharp-project-negative/services/UserService/Violations/CodeQuality/GeometryHelpers.cs` |

## architecture/deterministic/uri-property-as-string (#777)

OSS sources (OrchardCMS/OrchardCore):
- https://github.com/OrchardCMS/OrchardCore/blob/1d0ae14413f10e2b3c1a65927db463b30a6892fe/src/OrchardCore/OrchardCore.ContentFields.Core/Fields/LinkField.cs#L7
- ``https://github.com/OrchardCMS/OrchardCore/blob/1d0ae14413f10e2b3c1a65927db463b30a6892fe/src/OrchardCore/OrchardCore.ContentFields.Core/Settings/LinkFieldSettings.cs#L20``
[...]

PR #829, opened 2026-07-22
fix(fp): resolve 5 FPs from OrchardCMS/OrchardCore

DoilLee2222:
## Human review required

These fixes intentionally go **over the in-bounds boundary** (they touch modules outside `packages/analyzer/src/rules` + `patterns` + `tests/fixtures`), so `fp-next-fix-review` will review but **not** auto-merge — a human must review and merge.

Non-rule modules touched:
- **Roslyn semantic host** (`tools/csharp-roslyn-host/…`) — for #766 and #767 (`ExtensionMethodNamespace.cs`, `NamespaceFolderMismatch.cs`).
- **Data-flow module** (`packages/analyzer/src/data-flow/known-globals.ts`) — for #770.

#775 and #776 are in-bounds (rule visitors only) but ship in this human-review batch because the normal queue was empty.

Closes #766
Closes #767
Closes #770
Closes #775
Closes #776

## Fixes

| rule_key | issue | positive-fixture | negative-fixture |
|---|---|---|---|
| code-quality/deterministic/extension-method-namespace | #766 | unit: `roslyn-rules-code-quality.test.ts` (framework-namespace case) | existing `AccountStringExtensions.cs` + unit "flags same-namespace extension" |
| architecture/deterministic/namespace-folder-mismatch | #767 | unit: `roslyn-workspace.test.ts` (framework-namespace file) | existing `EmptyNamespaceResidue.cs` + unit "flags namespace ignoring its folder" |
| bugs/deterministic/no-undef | #770 | unit: `bugs-rules.test.ts` (`arguments`) | existing `legacy.js` + unit "detects undeclared variable" |
| performance/deterministic/event-listener-no-remove | #775 | `sample-js-project-positive/.../event-listener-lifecycle.ts` + unit | existing + unit "detects addEventListener without removeEventListener" |
| code-quality/deterministic/implicit-global-declaration | #776 | unit: `code-quality-rules.test.ts` (ES-module var) | existing `legacy.js` + unit "still flags var in a classic script" |

[...]

Issue #591, opened 2026-06-17
analyze crashes with "Invalid string length" on repos that vendor minified JS


DoilLee2222:
**Severity:** Medium (general analyzer-robustness). **NOT C#-specific** — surfaced on ServiceStack during PR #547 testing, but the snapshot bloat comes entirely from vendored **minified JavaScript**; it would equally affect any JS/TS repo that vendors minified assets. Pre-existing, independent of #547.

## Symptom
All analysis phases complete, then analyze dies at the save step, persisting nothing (`analyses/` empty, no `LATEST.json`):

```
● Saving results — Done
■  Invalid string length
```

## Root cause (confirmed by stack trace + heap probe)
`JSON.stringify(data, null, 2)` in `atomicWriteJson` (`packages/core/src/lib/atomic-write.ts`) throws `RangeError: Invalid string length` because the snapshot exceeds V8's max string length (~512 MB). On ServiceStack the `violations.added` array serialized to **709 MB**, ~90% of it the `snippet` field (chart.js alone ~496 MB; the largest single snippet was a 1.7 MB inline `//# sourceMappingURL=data:…base64,…` line).

Chain:
1. The repo vendors minified JS (`chart.js`, `swagger-ui.js`, `underscore-min.js`, …).
2. `looksMinified()` (`packages/analyzer/src/file-discovery.ts`) misses them: it only sniffs the first 8 KB and bails as "non-minified" on ≥4 newlines, so a leading multi-line license banner defeats it (`chart.js`). Separately, the `*.min.*` glob and `looksLikeBuildArtifact` don't match the **hyphen** form `underscore-min.js`.
3. The snippet builder (`packages/analyzer/src/rules/types.ts`) caps snippets at 3 lines but has **no character cap**, so one 166 KB+ minified line becomes a multi-hundred-KB snippet.
4. Snippets sum past 512 MB → `JSON.stringify` throws → analyze aborts before writing anything.

(abp, a *larger* repo with no giant minified JS, did **not** crash — confirming the crash is repo-content-specific, not size-driven.)

## Proposed fix (implemented & verified locally; #1 is highest-leverage)
1. **Cap snippet length** (~2 KB with an ellipsis) in `rules/types.ts` `makeViolation` (and the pyproject checker) — universal backstop that bounds the snapshot regardless of file type.
[...]

PR #833, opened 2026-07-23
fix(store): stream large snapshot serialization to avoid V8 string limit (#832 Blocker 1)

DoilLee2222:
## Summary

Fixes **Blocker 1 of #832**: whole-repo `analyze` crashes at the "Saving results" step on very large targets (e.g. `OrchardCMS/OrchardCore` — 95k+ violations across 5.5k files) with `RangeError: Invalid string length`, so no `LATEST.json`/snapshot is written and the campaign can't be re-measured. Same crash tracked in #829 / #831.

## Root cause

The store serialized every document with `JSON.stringify(data, null, 2)`, which must materialize the **entire** result as a single JavaScript string. V8 caps any string at ~512 MB (`buffer.constants.MAX_STRING_LENGTH` = 536,870,888 chars on 64-bit). A large analysis result serializes past that ceiling, so `JSON.stringify` throws before a byte is written.

The limit is symmetric on read: `JSON.parse(fs.readFileSync(file, 'utf-8'))` first builds one whole-file string, so even a file that large could not be read back to establish the next analyze baseline (`analyze-core.ts` reads `LATEST.json` on every run). Both sides had to be fixed.

The constraint is specifically the single-JS-string ceiling — a large object *tree* in memory is fine. So the fix is to never hold the whole document as one string.

## Fix

New `packages/core/src/lib/streaming-json.ts`:

- **`writeJsonStreaming`** — walks the value and streams it to the temp file incrementally (backpressure-aware, with a single persistent error listener so nothing leaks on a large document). Every leaf scalar is serialized by `JSON.stringify`, so unicode/escaping/number formatting is **byte-identical** to `JSON.stringify(x, null, 2)` — committed files (`LATEST.json`, `corpus.json`) don't reformat. Keeps the existing tmp-write + `rename` atomicity and removes the temp file on error. The largest string ever allocated is one leaf value.
- **`readJsonStreaming`** — parses the file as a byte stream via `stream-json` and assembles the object without ever concatenating the file into one string.

`analysis-store.ts` routes the large documents — `LATEST.json`, `analyses/*.json`, `diff.json` — through the streaming path (read + write). Small documents (`history.json`) keep the synchronous `atomicWriteJson`.

## Verification

- **Crossover proof:** a **626 MB** document that `JSON.stringify` cannot even produce (`throws RangeError: Invalid string length`) now writes and round-trips correctly through the streaming path.
- **New tests** (`tests/core/streaming-json.test.ts`): byte-identity vs `JSON.stringify(x, null, 2)` across a fixed edge-case corpus **plus 200 randomized structures** (unicode, string escaping, non-finite numbers → null, omitted `undefined`/function/symbol values, array holes, integer-like key ordering, `toJSON`), round-trip, reader parity with files written by plain `JSON.stringify`, atomicity on failure (target untouched, temp files cleaned up), and a multi-MB scale test exercising many flush cycles. An opt-in true >512 MB test is gated behind `TRUECOURSE_HUGE_JSON_TEST=1`.
- **No regressions:** `tests/core` green; the CLI `analyze` e2e ("writes a complete store and populates the registry") passes for JS/TS and Python, exercising the real `writeAnalysis`/`writeLatest`/`readLatest` round-trip through the streaming code.

## Scope notes

- Only the OSS file store is changed. The `guard-runner` and `ee` stores keep their own `atomicWriteJson` copies (not the reported crash); worth a follow-up if guard runs ever reach this scale.
- This does **not** address Blocker 2 (the six rule-contract disputes), which need human design decisions.

Fixes Blocker 1 of #832. Refs #829, #831.


DoilLee2222:
Keeping this open through the fp-automation retirement — flagging why, so it isn't swept up with the campaign issues.

Its original tracker #832 (`[cs-fp-campaign-stuck] OrchardCMS/OrchardCore`) is now closed as not planned along with the rest of the program. But the bug this fixes is **not** campaign scaffolding: `RangeError: Invalid string length` when the store serializes a large snapshot is product code in `analysis-store`, and **#591** still tracks it openly (repos that vendor minified JS hit the same V8 string limit without any campaign involved).

So the tracker for this work is #591, not #832. The OrchardCore repo was just the biggest thing we happened to point at.

PR #843, opened 2026-08-04
fix(guard): field-testing fixes — log boundary, setup transport parity + live phases, runners actually run, daemon-proof executor

### LINKED ISSUE #858: guard generate never constructs the fidelity/triage runners — 0 calls across 258, every failure commits untriaged
Field evidence (TrueCourse repo, v0.8.0-next.0, two full generates):

- LLM call logs show stage counts: `extract 38 · flows 28+1 · match 100+1 · generate 57+3 · retry 29+1` — and **zero `guard.fidelity` calls, zero `guard.triage` calls** across 258 total.
- Consequently all 18 committed-failing scenarios carry `diagnosis.triage: absent` — the dashboard verdict card has nothing to show, and the birth-failure routing degenerates: with no verdict, every failure takes the fail-open 'commit untriaged' path. Six scenarios that fail for one trivial authoring mistake (running `truecourse analyze` without `--llm/--no-llm`, aborting before the behavior under test) were committed as red 'drift'. The auto-resolve loop (retire → taint → re-author for generation defects) can never fire, and weak-but-green tests are never fidelity-reviewed.
- The estimate still prices both stages (fidelity 150 calls, triage 0–150 · $22.68 ceiling) — charged for stages that cannot run.

This is the exact bug shape the battle-test era hit and documented ('fidelity + triage runners always spawn — a runner that defaults to *not configured* silently disables a whole stage'); the reconciliation marked that fix obsolete-as-code with an explicit instruction to check the flows line's wiring for the same shape. The wiring check evidently failed: the in-process driver (`packages/core/src/commands/guard-in-process.ts` runner construction) appears not to construct/pass `fidelityRunner`/`triageRunner`, so the engine's 'no runner → fail-soft' branches swallow both stages silently.

Fix should include: construct both runners unconditionally in the driver; a regression test asserting a birth-failing generate produces triage verdicts end-to-end through the CLI driver (not just engine-level with injected runners — that's how this escaped: engine tests inject runners, the driver never builds them); and consider making 'stage configured but runner absent' loud rather than fail-soft.

mushgev:
Seven guard fixes from 0.8.0 field testing on the TrueCourse repo itself.

## 1. The `sinceLastStep` log window is scheduling-independent

CI proved the window marker racy under load. The marker now goes through the drain flush barrier: every byte the child handed to the OS before a step begins settles deterministically on the earlier side. Deterministic repro (red on old code), racy test expectation rewritten, 20× stress clean.

## 2. `guard setup` respects `llm.transport: api`

Field failure: setup spawned `claude --model gpt-5.5` (split config read: model from api config, transport not). Now mirrors `guard generate`: `preflightLlmOrExit`, shared `--llm-transport cli|agent|api` flag, one config read. Pinned: an api-configured model never reaches the claude argv; api-mode setup completes with no `claude` binary present.

## 3. Setup streams its verification phases

"Deriving the recipe" narrates live — `build 1m 4s · verifying: entry probe`, revision loops named; seed drafting likewise. Counters, never bars; lines written only on real transitions.

## 4. Fidelity + triage actually run (#858)

Field evidence: zero `guard.fidelity`/`guard.triage` calls across 258 in two full generates — every failing test committed untriaged, the auto-resolve loop dead. Root cause: both runners were gated on the CALLER's transport option, which is `undefined` by design in every default OSS run (the engine materializes its own default). The battle-test era fixed this exact bug with a warning comment; the flows rebuild restored the gate. Both runners now spawn unconditionally on the audited transport (accounting intact); a caller that genuinely can't reach a model aborts `llm-failed` BEFORE any write instead of silently committing an unadjudicated batch. Regression: an e2e through the real CLI driver with a fake `claude` binary and NO injected runners — injection is exactly how this escaped.

## 5. A spawned daemon can no longer hang the whole run

Field failure (the worst kind): a 60-scenario generate deadlocked forever mid-birth — zero CPU, no children, spinner frozen. A scenario ran `truecourse dashboard`, whose CLI spawns the dashboard server with `stdio: "inherit"`; the step's kill timer reaped only the direct child, the server survived holding the pipe write-ends, and the executor's close-only settle never fired. One stuck step parked the entire run behind the birth pool's `Promise.all`.

The executor now settles on the first of `close` / `exit` + a 1s stdio-drain grace / `error`, destroys its read ends on settle, and spawns steps detached with a process-group kill (mirroring the build path) so simple daemons are reaped instead of leaked onto the host. Orphaned stdio surfaces as an explicit infra reason — "the step left a background process still holding its output" — instead of a silent hang. Birth rounds get a wall-clock backstop (floor + per-candidate) applied inside `birthValidate` so no call site can forget it. Regression: an executor-level repro (child forks a `stdio: 'inherit'` grandchild and exits — hangs forever on the old code) plus a run-level case through a realistic `relkit watch` fixture command.

Because group-led children no longer share the terminal's foreground group, the CLI also installs a death sweep: on SIGINT/SIGTERM/SIGHUP or plain exit it SIGKILLs every live child process group (steps, builds, seeds, the api server) before re-raising, so a Ctrl-C mid-run kills everything guard spawned instead of orphaning it. Real-subprocess tests cover all three signals plus the exit path, each falsified with the sweep disabled.

## 6. The drain barrier stops guessing; the stall clock stops lying

The publish runner (slowest machine in the fleet) failed the pipe-buffer drain test 3/3 while every fast runner passed: stderr on a pipe drains at the mercy of the child's scheduling, and measurement showed NO parent-side timing window survives 12× oversubscription. The api server's stdio now lands in unlinked capture files — a write completes in the syscall, so everything written before a response is readable the moment the response lands, under any load. `drain()` keeps its contract and call sites as a forward read, and reading can no longer backpressure the system under test. New regression: a burst larger than the OS pipe buffer is complete the moment the response lands, with nothing to wait on. Separately, the cli transport's stall killer now verifies silence against its own measurement clock and re-arms for the remainder — it could previously kill a call at 79ms of an 80ms budget (cross-clock truncation).

## 7. The model follows the engine actually used (review: @spiderhands)

`effectiveLlmMode(flag)` is now the one answer to "which mode is this run in", threaded through model resolution, pre-flight estimates, and live model tags. Closes the reviewer's finding (saved api + `--llm-transport cli` → `claude --model gpt-5.5`) and its reverse (saved cli + `--llm-transport api` → no configured model) for guard setup / guard generate / spec scan; `analyze` is immune (never reads the api config) and pinned so it stays that way. Also aligns the two remaining stale `sinceLastStep` statements (authoring prompt, describe string) + a README instance; rolls the api author-prompt fingerprint (one-time api-surface re-author on next generate, cli cache unaffected).

Note for `chore/node-22` (#845/#848): its time-based drain rewrite conflicts with the file-capture `server.ts` — resolve by taking this branch's version; the `TC_LOG_RELEASE_MS` fixture knob becomes unnecessary but harmless.

Fixes #858.


spiderhands:
Two things I'd love to see closed before merge, both in the spirit of what the PR already argues for.

### 1. `--llm-transport cli` can still hand an API model to `claude`

`resolveModel`'s step 4 goes through `apiModeModel()`, which gates only on the **saved** mode:

```ts
// packages/core/src/config/global-config.ts:146
export function apiModeModel(): string | null {
  if (getConfiguredLlmMode() !== 'api') return null;
  return readApiLlmConfig()?.model?.trim() || null;
}
```

So with `llm.transport: api` + `model: gpt-5.5` saved, `truecourse guard setup --llm-transport cli` resolves the transport to `cliTransport()` while the model still resolves to `gpt-5.5` → `claude --model gpt-5.5`, which is the same deterministic exit 1 this PR fixes, just reached from the other direction.

To be clear, this isn't introduced here — `spec-in-process.ts` has the identical `resolveTransport` shape, so it's pre-existing across `spec scan` and `guard generate` too. But this PR adds the first explicit `cli` branch, which is the first time a per-run flag can actively contradict the saved selection, and the commit message's own rule ("one config, read once, or not at all") is what makes it feel worth finishing. Threading the effective transport into model resolution (an `apiMode` override on `resolveModel`/`resolveFallbackModel`) would close it once for all three commands.

### 2. Two other statements of the `sinceLastStep` semantic are now stale

`packages/shared/src/guard/scenario.ts` got the new wording, but the same rule is stated in two other places that still describe the old one:

- **`packages/guard-generator/src/prompts.ts:532`** — "`sinceLastStep` narrows the window to what the step before it produced". This is the LLM-facing authoring contract, and the commit message says the new attribution is "the idiom the generator prompt already teaches" — but the prompt never actually states the rule that `api-lifecycle.test.ts` had to be **inverted** for (a line belongs to the window that *opens* at the step that caused it, not to a later one). Left as-is, generate can keep authoring against the semantic that just changed.
- **`packages/shared/src/guard/describe.ts:348`** — `'only output written since the previous step counts'` renders to users; probably wants "since the previous step **began**".

mushgev:
Both closed in 01bf9dc4 — thanks for reading the resolution code instead of trusting the commit message.

**1.** Went with your threading suggestion, one level up: `effectiveLlmMode(flag)` (in `global-config.ts`) is now the single answer to "which mode is this run in" — an explicit flag beats the env override beats the saved selection — and it feeds model resolution, the pre-flight estimates (so the quoted price matches the models the run will actually use), and the live model tags. Audit of the flag's four commands: guard setup / guard generate / spec scan all had the hole and are fixed + pinned in both directions through the real transports (saved api + `--llm-transport cli` → no api model in the claude argv; saved cli + `--llm-transport api` → the api transport gets the configured model — the reverse cell was broken too). `analyze` turned out immune: it never consults the api config (`CLAUDE_CODE_MODEL` is its only model source), now pinned by test so it stays that way.

**2.** Both statements aligned with the window-opens-where-the-previous-step-began rule, plus a third instance you didn't cite in `README.md:392`. Heads-up on a side effect: the prompt edit rolls `GENERATE_API_PROMPT_FINGERPRINT`, so the first `guard generate` after this re-authors api-surface flows once (cli fingerprint unchanged, cli scenarios stay cached).

spiderhands (approved):

PR #853, opened 2026-08-04
perf(ci): cut PR feedback from ~14 min to ~3 min

DoilLee2222:
## Problem

A PR waited ~13.6 min for a verdict; a release took 16–38 min. Step timings from a green run showed why:

| Step | Time |
|---|---|
| checkout + setup node/pnpm/dotnet | 14s |
| `pnpm install` | 13s |
| `pnpm build` | 54s |
| `dotnet build` Roslyn host | 12s |
| **`pnpm test`** | **722s** |

The test step was 87% of the run, so caching alone could never fix it. Full design and measurements: `docs/CI_SPEED_PLAN.md`.

## Root cause: one .NET process per assertion

`tests/analyzer/roslyn-rules-*.test.ts` called `runRoslynHost()` inside every `it()`, and the client does `spawn('dotnet', …)` per call — ~450 process boots at ~0.8s each, ≈37% of all test CPU.

The host already reads requests in a loop until stdin closes and answers each with one response line (`Program.cs:51`), so the fix is to **hold one process open**, not to merge the snippets into a single request. Merging would have put every snippet into one Roslyn compilation and changed what the rules see; a session keeps each snippet its own `analyze` request and its own compilation, so results are identical.

- `openRoslynHost()` returns a session; `runRoslynHost` / `runRoslynWorkspace` are now one-shot wrappers over it, so there is a single code path.
- `useRoslynHost()` (`tests/analyzer/helpers.ts`) gives each rule file one session, closed in `afterAll`, opened lazily so a skipped file never spawns.

**Measured on the two heaviest files: 282s → 2.41s, all 380 assertions passing.**

## Also in this PR

- **Sharding** — both workflows run `vitest --shard=i/4`. The divisor is `strategy.job-total`, so the matrix list is the only thing to edit. `tests-passed` aggregates the shards into one stable check.
- **Shared setup** — `.github/actions/setup` holds the toolchain, caches and builds for both workflows so they cannot drift. Adds pnpm store, `.turbo` and NuGet caches; CI was logging `Cached: 0 cached, 20 total` on every run.
- **Narrower build** — `pnpm build:test` skips the landing and dashboard-client apps, whose build output no test imports. Both already had `typecheck` scripts, now wired to a `typecheck` job that runs alongside the shards, so no type coverage is lost.
- **Cancel superseded runs** — 10 of the last 60 runs overlapped on the same branch, wasting 87 min.

## publish.yml still runs the whole suite — on purpose

It looks redundant. It is not: release tags sit on commits `test.yml` never sees. All six `v0.7.4-*` tags point at commits **not on `main`**, and a search of the last 100 runs finds no run for `5a75baac`. That publish run caught a real defect and blocked the release:

```
FAIL tests/core/corpus-in-process.test.ts > curateInProcess
LlmStageFailureError: every LLM call in the `spec.vocab` stage failed (1 of 1)
  — First failure: spawn /nonexistent/claude-test-tripwire ENOENT
```

So it is sharded, never skipped. Its jobs are now `gate → meta → test → publish`, where `meta` resolves the version, tag and commit once and every later job checks out that SHA — the tested commit is provably the published one. Tagging moved after the tests, so a failing suite no longer leaves a tag behind.

Node versions stay split: PRs on 20 (the `engines` floor and the esbuild target), releases on 22 (what most users run).

## Verification

- Two heaviest Roslyn files: 380 tests pass in 2.41s (was 282s).
- Full suite run after deleting `apps/dashboard/client/dist` and `apps/landing/build` and building with `build:test` only — nothing needed them.
- All four shards run: 123 + 123 + 123 + 123 = 492 files, matching the unsharded suite.
- `pnpm typecheck` green, covering both apps.

**Known, pre-existing:** the full suite flakes on ~1 random file per local run. Confirmed unrelated — pristine `main` does it too (`split-analyzer-layers.test.ts`), each run picking a different file, and every one passes in isolation. Not addressed here; it deserves its own investigation.

## Follow-up

`main` is currently unprotected, so nothing outside the repo had to change. If protection is enabled later, require **`tests-passed`** rather than the individual `test (1..4)` checks, so the shard count can change without touching repository settings.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,71 @@
+name: Set up the build
+description: >-
+  Toolchain, caches, dependencies and the builds every CI job needs before it can
+  run the test suite. Shared by test.yml and publish.yml so the two can never
+  drift apart.
+
+inputs:
+  node-version:
+    description: >-
+      Node major to run on. PRs run the `engines` floor (20, which is also the
+      esbuild target); releases run the version most users are on (22).
+    required: true
+  registry-url:
+    description: >-
+      npm registry to write into `.npmrc`. Only the publish job needs it; leaving
+      it empty keeps the runner's default registry configuration.
+    required: false
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - uses: pnpm/action-setup@v4
+
+    # `cache: pnpm` needs pnpm on PATH, so this must follow action-setup.
+    - uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+        registry-url: ${{ inputs.registry-url }}
+        cache: pnpm
+
+    # The C# semantic rules run in an out-of-process .NET Roslyn host. Without
+    # the SDK + a built host, every C# host test skips and analyze-csharp fails
+    # (the pipeline is host-required for C#). Build it so C# is actually tested.
+    - uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: '8.0.x'
+
+    # Turbo's local cache. Without it every run rebuilds all 20 packages from
+    # scratch — CI logs showed `Cached: 0 cached, 20 total` on every single run.
+    # The per-SHA key always misses; the prefix restore-key supplies the previous
+    # run's cache, from which turbo replays the packages that didn't change.
+    - uses: actions/cache@v4
+      with:
+        path: .turbo/cache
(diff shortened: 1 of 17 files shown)


DoilLee2222:
Superseded — reopening without the build-scope change (item 3 in docs/CI_SPEED_PLAN.md), which is deferred as documented.

PR #879, opened 2026-08-21
feat(integrations): reveal a stored token for recovery (TEMPORARY)

mushgev:
> **Temporary.** This exists to recover the Confluence and Jira tokens that only exist as ciphertext. Revert it once they have been re-entered.

## Why

`integration_connections.token_enc` is AES-256-GCM keyed on `sha256(TRUECOURSE_SECRET_KEY)`, and that secret is deliberately distinct per environment (`infra/azure/README.md`). A row copied between environments carries ciphertext the receiving side cannot authenticate, and there is no way to read the original value back out.

## What

- `POST /api/ee/integrations/:kind/reveal` returns the plaintext token for the caller's own workspace.
- `IntegrationStore.revealToken()`, separate from `getConnection()` so the one caller that deliberately surfaces a secret stays greppable, and so a key mismatch is a reportable state rather than a throw.
- Reveal button in the Integrations drawer, plaintext in a `select-all` block with Hide.

No infra or workflow changes: `containerapp.bicep` and `deploy-prod.yml` are identical to `main`. Deploying is the normal Deploy (prod) run.

## Behaviour

- Scoped to the caller's workspace through the existing `eeUser.organizationId`.
- The token is never logged. The audit line carries connector, org and user only, and nothing goes to Sentry.
- `Cache-Control: no-store`; the plaintext lives in component state that dies with the drawer.
- A row encrypted under a different master secret answers **409** naming the real cause, rather than leaking or crashing.
- No stored token 404s; no workspace 401s.

## Tests

3 cases in `tests/ee-server/integrations-route.test.ts`: plaintext + `no-store`, the 409 foreign-key path, and the no-token / no-org paths. Affected suites green: 101 files, 1304 tests.

## Not included

The `/test` route's 502: `store.getConnection()` sits outside the handler's try block in `integrations/index.ts`, so a decrypt failure escapes as a 502 instead of a clean 400. Left alone deliberately, separate fix.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -7,6 +7,11 @@ name: Deploy (prod)
 # Provision the prod RG + configure the prod Environment first (see infra/azure/README.md).
 on:
   workflow_dispatch:
+    inputs:
+      allow_token_reveal:
+        description: 'TEMPORARY: allow revealing stored integration tokens in plaintext. Leave false unless recovering a token, and redeploy with it false afterwards.'
+        type: boolean
+        default: false
 
 permissions:
   id-token: write
@@ -59,4 +64,5 @@ jobs:
                environmentId="${{ steps.infra.outputs.env_id }}" \
                identityId="${{ steps.infra.outputs.identity_id }}" \
                acrLoginServer="${{ steps.infra.outputs.acr_login }}" \
-               keyVaultUri="${{ steps.infra.outputs.kv_uri }}"
+               keyVaultUri="${{ steps.infra.outputs.kv_uri }}" \
+               allowTokenReveal="${{ inputs.allow_token_reveal }}"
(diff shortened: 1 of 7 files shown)


mushgev:
Temporary, and no longer present in any environment.

This existed to recover a Confluence/Jira token that only existed as ciphertext. `integration_connections.token_enc` is AES-256-GCM keyed on a per-environment `TRUECOURSE_SECRET_KEY`, so a connection row copied between environments cannot be decrypted on the receiving side and the original value cannot be read back out.

Status:

- Reverted out of `main`. `main` is back at `eafab32a`, and the route is not in the current tree.
- Prod redeployed without it, now running `main-eafab32`.
- The feature branch has been deleted.

The route was scoped to the caller's own workspace, never logged the token, sent `Cache-Control: no-store`, and returned 409 rather than leaking when a row was encrypted under a different master secret.

PR #896, opened 2026-09-10
feat(dashboard): the Agent page lists every run of the workspace and opens each as its verbatim conversation

mushgev:
## What

**Activity leaves the repository console.** A top-level **Agent** page (`/preview/agent`) lists the workspace's runs across its connected repositories, newest first, behind one filter row (Kind, Status, Repository), with `?repo=` deep links from the console, Home and the job toasts. The console's Activity tab and the mock activity views are gone.

**A run opens as one conversation rendered from its stored transcript** (`/preview/agent/:runId`): the run's steps as headings with their facts, and under each the messages of its inner sessions in the order they were written.

- the system prompt and the briefing as folded text, opening in place
- every model turn as its tool call with the result that answered it, paired even when the loop's own events (a child the tool spawned, a grant) came between them; failed calls red on the name and the result
- the outcome value as JSON, failures and the loop's events as their own fields
- the client paraphrases nothing: the narration fold, tool phrases, per-kind vocabulary and outcome digests are deleted, and a test guards against their return
- the outline marks the step the reader is in; the page opens at its end, follows a live run, and offers a jump back down; a message box waits at the bottom, disabled until a route can deliver to a live session

**Server**

- `GET /api/sessions/runs?repo=&kind=&status=&limit=&before=` lists runs workspace-wide with keyset paging; `GET /api/sessions/runs/:runId` reads one
- `GET /api/repos/:id/sessions/runs/:command/:runId/activity?after=&limit=` pages a run's journal
- the dead-run sweep runs once per listing across repositories
- `FileLogTransport` absorbs a failed sink instead of raising an unhandled error; the workspace LLM test waits for the analyze it started before tearing its fixture down

**Also**: every preview page renders its header through `PageHeader`, which learned a crumb trail; the console's trail starts at Home.

## Verified

- typecheck clean (client, server)
- client project 65 files / 967 tests; node project over server, data-store, core, jobs, agent-loop, architecture 137 files / 1684 tests
- the workspace listing and journal paging exercised read-only against a restored production dump (3 repositories, 40 runs, 34k journal events)
- rendering checked in the browser on the expense-tracker spec scan and guard generate runs


spiderhands (approved):
Nicely done! 💪

